PR #59, opened 2020-07-30
AllDifferent constraint

bilelomrani1:
Here is the first version of the Arc-consistent `AllDifferent` constraint, based on  "A filtering algorithm for constraints of difference in CSPs" J-C. Régin, AAAI-94. Not fully optimized yet (I need to understand better `freeIds` mechanisms) and more tests are needed.


qcappart:
I close this PR as another AllDifferent implementation has been proposed and merged.

PR #65, opened 2020-07-30
Rename and make Lexicographic the default heuristic

### LINKED ISSUE #63: Make Lexicographic the default value of a BasicHeuristic
I think that the `Lexicographic` value ordering `BasicHeuristic(x -> minimum(x.domain))`  is more common in CP than `BasicHeuristic(x -> maximum(x.domain))`. The naming is also more explicit. I propose simply replacing `BasicHeuristic` by a lexicographic ordering. What do you think?

Because this can cause problems of reproducibility in your personal experiments, I can wait your green light.

bilelomrani1:
Fixes #63


qcappart:
Outdated with the current SeaPearl version

Issue #49, opened 2020-07-24
Add the all different constraint 


felixchalumeau:
All different is a very common constraint in CP. It would thus be cool to have it in the solver !

PR #115, opened 2021-04-26
Feature/alldifferent

PierreTsr:
This is my proposition for the _AllDifferent_ constraint, relating to the issue #49. It contains:

- The AllDifferent constraint;
- The associated testset;
- The supporting algorithm for matching computation;

My implementation uses only `LightGraphs.jl` and no third-party solver. 
The downside is that I manually implemented the matching algorithm and thus introduced many potential inefficiency sources. I tried to use `LightGraphsMatching.jl` but the BlossomV modules breaks the cross-platform portability as it requires some C++ files. 
Moreover the matching algorithm code is placed in the same file as the AllDifferent constraint as no dedicated folder exists for backend algorithms currently. But it could be wiser and cleaner to move it to a dedicated part of the architecture if more complex constraints are to be added in the future.

Thus potential improvements are:

- [x] Adding a function to complete an existing matching instead of going back from scratch each time a new matching is needed;
- [x] Moving the matching algorithm to another part of the project;
- [x] Adding more complex test cases for intermediate functions;
- [x] Increasing code readability;
- [ ] Replacing the algorithm with a more recent one;
- [ ] Using benchmarking tools to track unefficient code;

One hesitation I had is in the last 2 test: I created a n-queens model to grant a full test of the constraint. But these tests thus rely on many other parts of the code. Is this anti-pattern? Because otherwise it is hard to guarantee the correct behavior of the constraint (for instance, I missed a few solutions in the n-queens at some point, but solving the problem integrally was the only way to detect this bug).


PierreTsr:
has an inconsistant behavior been introduced recently in the tsptw.jl test? Because it ran smoothly until I merged the last version of `master` and now the test seems to fail only part of the time. I haven't modified any of the related files afaik

kimriouxparadis (commented):
Overall great code! 
I would like to discuss the variable and function names, but otherwise it seems good!

kimriouxparadis (approved):

PR #117, opened 2021-05-06
Fix for Absolute.jl

PierreTsr:
I detected 2 bug in the new constraint `Absolute`

1. The function `propagate!` wasn't checking for the feasability of an assignment before enforcing it, and thus triggered errors instead of returning `false`;
2. The function `propagate!` wasn't removing the constraint itself from `toPropagate`, which lead to infinite loops of constraint propagation.

I propose fixes for these bugs, and added some test cases to ensure they were properly fixed. Moreover I added a definition of `show` for the constraint to display it nicely. Finally, I fixed a missing `include` in the tests.

### DIFF
--- a/src/CP/constraints/absolute.jl
+++ b/src/CP/constraints/absolute.jl
@@ -26,15 +26,24 @@ end
 function propagate!(constraint::Absolute, toPropagate::Set{Constraint}, prunedDomains::CPModification)
 
     if isbound(constraint.x) 
-        prunedY = assign!(constraint.y, abs(assignedValue(constraint.x)))
+        value = abs(assignedValue(constraint.x))
+        if !(value in constraint.y.domain)
+            return false
+        end
+
+        prunedY = assign!(constraint.y, value)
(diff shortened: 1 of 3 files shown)


src/CP/constraints/absolute.jl
```diff
@@ -26,15 +26,24 @@ end
...
 
     elseif isbound(constraint.y) # x is equal to y or -y
+        value = assignedValue(constraint.y)
+        if !(value in constraint.x.domain || -value in constraint.x.domain)
+            return false
+        end
     
         for v in minimum(constraint.x.domain):maximum(constraint.x.domain)
-            if v != assignedValue(constraint.y) && v != -assignedValue(constraint.y)
+            if v != value && v != -value
```
PierreTsr, line 46: On second thought, this case isn't necessary as it doesn't trigger errors and will be handled at the end if `x.domain` becomes empty. Should we keep it for symmetry though ?

qcappart (approved):
Good work, seems good and complete to me

PR #119, opened 2021-05-10
Table constraint: compact table

PierreTsr:
This my implementation proposition for the CompactTable algorithm. I followed the implementation from Demeulenaere J. et al. (2016) 
Compact-Table: Efficiently Filtering Table Constraints with Reversible Sparse Bit-Sets. [https://doi.org/10.1007/978-3-319-44953-1_14](https://doi.org/10.1007/978-3-319-44953-1_14)

The PR contains:

- The Reversible Sparse BitSet datastructure;
- The `TableConstraint` constraint;
- Both dedicated testset, coauthored by @tomsander1998 ;
- Small changes in the `expandDFS!` and `fixpoint!` functions;

The changes I propose in `expandDFS!` and `fixpoint!` functions should get a particular attention as they are in the core of the CPModel. They are fixing a behavior I consider misleading, and was particularly annoying in the `TableConstraint`: 
the variable `prunedDomains::CPModification` given to `propagate!` seems currently useless, as it doesn't contain the values pruned during the branching (i.e. values removed by `assign!` or `remove!`) and thus cannot be reliably trusted to get the domains modifications. Moreover, as its final value doesn't seem to be used anywhere, I assume that its goal is to give a fast access to current domain changes in `propagate!`.
Thus, I modified both `expandDFS!` and `fixpoint!` functions to initialize `prunedDomains` with the values removed during branching.

The only difference between my code and the paper is during the update of `modifiedVariables`. In SeaPearl the domain-length cannot be trusted to identify modified variables (when branching on a variable with 2 values, both branches will have a 1-sized domain with different values, and the given code won't detect it). I used the corrected `prunedDomains` to detect these cases, and the constraint should now detect all the modified variables.

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,0 +1,15 @@
+name: TagBot
+on:
+  issue_comment:
+    types:
+      - created
+  workflow_dispatch:
+jobs:
+  TagBot:
+    if: github.event_name == 'workflow_dispatch' || github.actor == 'JuliaTagBot'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: JuliaRegistries/TagBot@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.DOCUMENTER_KEY }}
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .DS_Store
 /Manifest.toml
 /dev/
+.vscode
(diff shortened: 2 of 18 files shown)


qcappart (approved):
Excellent work ! Implementation seems consistent with the paper (excepting for the incremental/reset-based update).

I approve the modification but we should double check with Ilan and Félix for the change in ExpandDFS

kimriouxparadis (approved):

PR #129, opened 2021-05-17
fix: SetDiffSingleton propagation

### LINKED ISSUE #122: Bug in the TSPTW test
An inconsistant behavior has been spotted on the test `Search known instance` line 166 in `test/datagen/tsptw.jl`.

The test aims at checking the correct generation of tsptw instances, by solving already resolved problems and controling the existence of a solution. However, this test randomly fails despite the specification of a seed, which make the construction of the instance deterministic.

It appears that the issue doesn't come from the problem generation, but from the CP solver itself. After a quick investigation, I managed to find the point where the bug occurs in the search tree. Indeed, as both heuristics are deterministic in this test, the solver always explores the same path in the search tree. But at some point, a variable assignment triggers a random behavior in `fixpoint!` resulting in a valid or invalid assignment without any obvious explanation.

I tweaked the original test a little bit to reproduce the bug:
``` julia
using SeaPearl

trailer = SeaPearl.Trailer()
model = SeaPearl.CPModel(trailer)

n_city = 21
grid_size = 150
max_tw_gap = 3
max_tw = 8

generator = SeaPearl.TsptwGenerator(n_city, grid_size, max_tw_gap, max_tw)

SeaPearl.fill_with_generator!(model, generator; seed=42)

SeaPearl.assign!(model.variables["a_1"], 19)
SeaPearl.assign!(model.variables["a_2"], 20)
SeaPearl.assign!(model.variables["a_3"], 16)
SeaPearl.assign!(model.variables["a_4"], 11)
SeaPearl.assign!(model.variables["a_5"], 14)
SeaPearl.assign!(model.variables["a_6"], 8)
SeaPearl.assign!(model.variables["a_7"], 17)

SeaPearl.fixPoint!(model)

SeaPearl.assign!(model.variables["a_8"], 5)

status, prunedDomains = SeaPearl.fixPoint!(model, SeaPearl.getOnDomainChange(model.variables["a_8"]))
```
On my system on successive runs, `fixPoint!` returns `false` a third of the time.
On every failed run I've had in the original test, this last assignment always was the one triggering the failure.

I double-checked that the distance matrix and time_windows generation was consistant (supposedly the only variables randomly generated in `TsptwGenerator`). I haven't been able to check that the initialization of every variable/constraint is the same at each run, but they aren't suppose to be random. I haven't been able to check yet which constraint/part of `fixPoint!` is responsible for this behavior.

However, this part becomes tricky because there are 1500+ constraints in the model and `fixPoint!` doesn't seem to call them all in the same order. I'll keep digging and will post any new lead.

My current leads are:

- a failing constraint pruning to much values
- a mishandling of a variable in `fixPoint!`

PierreTsr:
This a fix for the `SetDiffSingleton` propagation error. I rearranged a bit the code which was a bit weird sometime, and added a `toRemove` buffer to avoid pruning values on the domain we are iterating on.

This PR should fix #122 .

### DIFF
--- a/src/CP/constraints/setdiffsingleton.jl
+++ b/src/CP/constraints/setdiffsingleton.jl
@@ -28,18 +28,16 @@ function propagate!(constraint::SetDiffSingleton, toPropagate::Set{Constraint}, 
     x = constraint.x
 
 
-    # For all b possible, exlucde if not possible in a and x not assigned to the value
+    # For all b possible, exclude if not possible in a and not possible for x
+    toRemove = Int[]
     for v in possible_not_required_values(b.domain)
-        if !is_possible(a.domain, v) && !(isbound(x) && assignedValue(x) == v)
-            if is_possible(b.domain, v)
-                if !is_required(b.domain, v)
-                    exclude!(b.domain, v)
-                    triggerDomainChange!(toPropagate, b)
-                else
-                    return false
-                end
-            end
+        if !is_possible(a.domain, v) && !(v in x.domain)
+            push!(toRemove, v)
         end
+    end
+    for v in toRemove
+        exclude!(b.domain, v)
+        triggerDomainChange!(toPropagate, b)
     end
 
     # For all b required, require in a if not in x domain, assign to x if not possible in a, remove from x domain if required in a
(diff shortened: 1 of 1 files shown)


qcappart (approved):
Sounds good to me! I tested my problematic seed (and few others) and all the tests were successfull. Minor suggestion: add all the set constraints, in a sub-directory (`constraints/set`) in order to make it easier to read when new constraints will be added.

PR #130, opened 2021-05-17
a prettier SeaPearl

PierreTsr:
During my recent debugs, I added a few `Base.show` dispatches for SeaPearl variables and constraints in order to get clearer insights on what was going on. I thought these features could be interesting to add to the main branch, as it could help others in the future.


qcappart (approved):
Great ! As only the `Base.show` has been changed, I can directly approve it. In the future, It would be nice to have a developer guide to formalize everything that should be done when adding a constraint/functionality. :-)

PR #131, opened 2021-05-17
Removed extra dependencies

PierreTsr:
Some extra dependencies have been recently added to SeaPearl by mistake, but aren't needed anywhere.
I propose to remove:

- Gadfly
- GraphPlot
- JuliaFormatter
- LanguageServer
- CUDA (it was installed before, but it is useless as long as we don't handle GPU computation. It will be needed later, but we shouldn't have it as a dependency for the time being, as it poses issues with other packages)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,18 +4,13 @@ authors = ["Ilan Coulon", "Félix Chalumeau"]
 version = "0.2.0"
 
 [deps]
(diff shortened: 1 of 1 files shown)


qcappart:
I guess you were annoyed by the multiple PR of compactHelper. :D But yes, I agree with  the removal

PierreTsr:
Haha that's part of it, but I discussed about it with Tom and he agreed he didn't do it on purpose ^^

PR #133, opened 2021-05-18
iLDS

3rdCore:
This is my proposition for the implementation of the iLDS Search strategy. One major challenge of implementing an Iterative Limited Discrepancy Search is that one can not know in advance the maximal discrepancy of some paths of the search tree in the context of CP programming. Indeed, at each step of the exploration, the strategy decides to follow the heuristic ( assign x to v ) or to not follow it ( remove v from x domain ). Hence, the total distance of the path depends on the successive fix-point along the path.  

I used what I called virtual _Discrepancy_ and _Depth_ to tackle this difficulty. The complexity is however worse than DFS as the algorithm needs to pass several times on the same node to find solutions on the subtree with different discrepancies K =0,1,2,3 .... 

**Key features :**
- generalization of the search function
- implementation of the iLDS strategy

### DIFF
--- a/src/CP/core/search/ilds.jl
+++ b/src/CP/core/search/ilds.jl
@@ -1,4 +1,12 @@
 
+
+
+"""
+initroot!(toCall::Stack{Function}, ::Type{DFSearch},model::CPModel, variableHeuristic::AbstractVariableSelection, valueSelection::ValueSelection, newConstraints=nothing)
+
+generic function to instantiate the research based on a specific Strategy <: SearchStrategy. The max discrepancy correspond to the number or branchable variables at 
+the beginning of the search. Calls to expandIlds! with a decreasing discrepancy is stacked in the toCall Stack. 
+"""
 function initroot!(toCall::Stack{Function}, ::Type{ILDSearch}, model::CPModel, variableHeuristic::AbstractVariableSelection, valueSelection::ValueSelection, newConstraints=nothing)
     isboundedlist = [!isbound(v) for (k,v) in model.variables]
     @assert !isempty(isboundedlist) "initialisation failed : no declared variables"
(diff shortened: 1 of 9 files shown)


qcappart (approved):

src/CP/core/search/ilds.jl
```diff
@@ -0,0 +1,77 @@
...
+"""
+function initroot!(toCall::Stack{Function}, ::Type{ILDSearch}, model::CPModel, variableHeuristic::AbstractVariableSelection, valueSelection::ValueSelection, newConstraints=nothing)
+    isboundedlist = [!isbound(v) for (k,v) in model.variables]
+    @assert !isempty(isboundedlist) "initialisation failed : no declared variables"
+    depth = sum(isboundedlist)
+    
+    # Note that toCall stack is a LIFO data structure, expandIlds with a discrepancy threshold of 0 will be the first one to execute (then with 1, 2, 3, etc.)
+    for k in depth:-1:1
+        push!(toCall, (model) -> (nothing;expandIlds!(toCall,k,depth, nothing, model, variableHeuristic, valueSelection)))
+    end
```
PierreTsr, line 18: I'm not sure to see why there is `nothing;` in the function

src/CP/core/search/ilds.jl
```diff
@@ -0,0 +1,77 @@
...
+        push!(toCall, (model) -> (restoreState!(model.trailer); :BackTracking))
+        push!(toCall, (model) -> (remove!(x.domain, v); expandIlds!(toCall,discrepancy, depth, :Right, model, variableHeuristic, valueSelection, getOnDomainChange(x))))
+        push!(toCall, (model) -> (saveState!(model.trailer); :SavingState))
+    end
+
+    if (depth>discrepancy)
+        push!(toCall, (model) -> (restoreState!(model.trailer); :BackTracking))
+        push!(toCall, (model) -> (assign!(x, v); expandIlds!(toCall,discrepancy, depth-1, :Left, model, variableHeuristic, valueSelection, getOnDomainChange(x))))
+        push!(toCall, (model) -> (saveState!(model.trailer); :SavingState))
+    end
```
PierreTsr, line 74: It would be better if you use the updated version of this code in `expandDfs!`. Otherwise `prunedDomains` won't be set correctly in the constraint propagation, and it would harm `tableCT`.

PierreTsr (commented):
I haven't check in details the logic of the implementation yet. I'll get back to it.

src/CP/core/search/ilds.jl
3rdCore: For generalization purposes, functions pushed in the stack needed to have the same formalism as the one stacked during the DFS process.

src/CP/core/search/ilds.jl
3rdCore: Okey thanks for the warning Pierre !

3rdCore: I just fixed it !

PR #135, opened 2021-05-18
all diff for arrays

tomsander1998:
Small modif from Vector to Array so that the constraint can take any Array as argument

### DIFF
--- a/src/CP/constraints/alldifferent.jl
+++ b/src/CP/constraints/alldifferent.jl
@@ -12,7 +12,7 @@ end
 
 AllDifferent constraint, enforcing ∀ i ≠ j ∈ ⟦1, length(x)⟧, x[i] ≠ x[j].
 
-test The implementation of this contraint is inspired by:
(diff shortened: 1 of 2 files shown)


src/CP/constraints/alldifferent.jl
```diff
@@ -8,17 +8,17 @@ include("algorithms/matching.jl")
 end
 
 """
-    AllDifferent(x::Vector{<:AbstractIntVar}, trailer)
+    AllDifferent(x::Array{<:AbstractIntVar}, trailer)
 
 AllDifferent constraint, enforcing ∀ i ≠ j ∈ ⟦1, length(x)⟧, x[i] ≠ x[j].
 
-The implementation of this contraint is inspired by:
+test The implementation of this contraint is inspired by:
```
PierreTsr, line 15: I don't think this should be here haha

PierreTsr (approved):
Good to me!

PR #138, opened 2021-05-20
Compact Table optimization

### LINKED ISSUE #137: Possible TableCT optimisation
After discussing the problem _Eternity 2_ with @tomsander1998 we figured out a potential optimisation of the table constraint.

Indeed, in _Eternity 2_, each piece is represented by a tuple of 5 variables which must obey a table constraint. However, each table constraint has the same table and thus the same supports.

With the current implementation, the table is copied each time, and the support is rebuilt each time which `#pieces` times more memory than if we had a single table/support dictionnary.

What should be done is defining a new constructor for `TableConstraint` that take a table & support dictionnary as input and set the corresponding attributes as references to these.

PierreTsr:
This is the new constructor proposed in #137 with the associated testset.

I first changed the function `buildResidues` as some supports weren't present in the table when initializing the constraint. But then I figured it would be more efficient to give the constraint already cleaned versions of the supports and table. 

The counterpart is that between different instances of the constraint sharing the same instances of `table`/`supports`, the variables should initially have the same domains (otherwise the support will be setup incorrectly for some instances).

To sum up, this new constructor gives much more freedom to the user but should be well understood in order not to tigger errors.

Closes #137

### DIFF
--- a/src/CP/constraints/compacttable.jl
+++ b/src/CP/constraints/compacttable.jl
@@ -3,12 +3,17 @@ include("algorithms/rsparsebitset.jl")
 """
     TableConstraint
 
-Constraint enforcing that: ∃ j ∈ ⟦1,m⟧, such that ∀ i ∈ ⟦1,n⟧ xᵢ=table[i,j].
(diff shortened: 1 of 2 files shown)


qcappart (approved):
Sounds good to me ! It will be interesting to access the memory reduction that you gain with this optimization

kimriouxparadis (approved):

PR #140, opened 2021-05-25
CPModification upgrade

### LINKED ISSUE #128: prunedDomains update check
During the `propagate!` function, the variable `prunedDomains` is supposed to store all the domain modifications. However, @ilancoulon  warned us that the update isn't entirely reliable yet. As it is a very useful feature, it would be great to ensure that this variable is always properly updated.

PierreTsr:
As proposed in #128, this PR brings new features to ensure the correctness of `prunedDomains` during each propagation. Thus te mains contributions are:

- the struct `SetModification` to make it possible to track changes in a set domain;
- a proof-read of every constraint to make sure that each domain change is followed by an `addToPrunedDomains!`;
- a testset dedicated to `CPModification`, to ensure that each constraint is updating it well.

During this work I encountered a few undetected bugs and made some minor changes:

- missing cases in `SetEqualConstant`;
- a few display error;
- prevent `Absolute` to prune values before the beginning of the execution, as this pruning is intractable for the other constraints.

If this PR is approved, we should now have a reliable `prunedDomains` variable, as long as all the new constraints follow a few guidelines. It would be great to write a few instructions about this for potential contributors.

Closes #128

### DIFF
--- a/src/CP/constraints/absolute.jl
+++ b/src/CP/constraints/absolute.jl
@@ -10,6 +10,7 @@ struct Absolute <: Constraint
 
     function Absolute(x, y, trailer)
         constraint = new(x, y, StateObject(true, trailer))
+        # TODO fix domain change
         removeBelow!(y.domain, 0) # y cannot take a negative value
         addOnDomainChange!(x, constraint)
         addOnDomainChange!(y, constraint)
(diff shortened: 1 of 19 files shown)


qcappart (commented):
Very useful PR ! :-) Thanks for this refactoring

kimriouxparadis (approved):

PR #139, opened 2021-05-21
Feature/disjunctive

kimriouxparadis:

PR #141, opened 2021-05-26
Entire refacto of search metrics 

3rdCore:
This is my proposition for the implementation of a generic and adaptable metrics to store many useful pieces of informations and results obtained during the search, such as the number of nodes visited for every solution for every episode during the training, the time needed to complete a search on an episode...

Previously, the metrics had to be defined by the user while creating his model. This was not relevant as the retrieved results don't depend on the kind of problem ( be it knapsack, graph coloring ... ) addressed by **SeaPearl**.   

By using the abstract struct : `AbstractMetrics` the user can define his own `CustomMetrics` as long as it satisfies some requirements. This PR also provides a `basicmetrics` that can be used _out of the box_.

This PR goes in hand with an another PR on SeaPearlZoo that will update the current examples (knapsack, graph_coloring, tsptw) to fit this new format. 

**Key features :** 
-  basicmetrics can be used on any kind of CPModel (containing an objective or not ) 
-  basicmetrics can be used on any kind of heuristic ( classic or trained ) 
-  basicmetrics provides useful plotting functions : ( evolution of nodes visited to find a first solution / prove the optimality along the training, evolution of score of solutions found compared to the optimal one during the search during an evaluation.

- Evolution of nodes visited to find the first solution / prove the optimality along the training : ( for illustration purpose, no noticeable learning can be observed )
Capture d’écran du 2021-05-26 18-30-33

- Evolution of score of solutions found compared to the optimal one during the search during an evaluation :
Capture d’écran du 2021-05-27 11-30-56


src/CP/core/search/search.jl
```diff
@@ -42,7 +42,7 @@ function search!(model::CPModel, ::Type{Strategy}, variableHeuristic::AbstractVa
     end
     
 
-    if length(model.solutions) > 0
+    if length(model.statistics.solutions) > 0
```
tomsander1998, line 45: this corresponds to what case ?

qcappart:
Seems to be a great enhancement ! 

Few questions/concerns:

- The checks have failed. I guess it is due to a mismatch with the previously accepted PR #139 
- I am not sure to understand the last picture. What are the different curves ?
- Also, I would be interesting to have a native support on the evolution of the best known solution so-far during the search. I.e., the number of nodes visited in the x axis, the the solution value on the y-axis.

qcappart:
Ok thanks ! I understand the last picture. We should find a way to reduce the number of curves. For instance, we can add a parameter defining each "how many" episode a new curve must be plotted.

Also, in practice, it may be difficult to have the relative score, as it requires to know the optimal solution. Maybe we can change it with the relative score to the best solution found. Also, it could be interesting to have a measure of the "raw value" of the objective.

src/experiment/metrics/metrics.jl
```diff
@@ -0,0 +1,40 @@
+
+
+
+"""
+        AbstractMetrics
+
+AbstractMetrics is a data structure that allows the user the stock relevant informations 
+of consecutive search along learning (either on different instances or on the same one).
```
PierreTsr, line 8: data structure => abstract type
informations => information
same one => same ones

src/experiment/metrics/metrics.jl
```diff
@@ -0,0 +1,40 @@
+
+
+
+"""
+        AbstractMetrics
+
+AbstractMetrics is a data structure that allows the user the stock relevant informations 
+of consecutive search along learning (either on different instances or on the same one).  
+
+SeaPearl allows the user to define his own CustomMetrics as long as the it contains :
```
PierreTsr, line 10: as the it => as it

src/experiment/evaluation.jl
```diff
@@ -2,42 +2,41 @@ abstract type AbstractEvaluator end
...
-function SameInstancesEvaluator(; eval_freq = 50, nb_instances = 50)
-    SameInstancesEvaluator(nothing, eval_freq, nb_instances)
-end
-
-function init_evaluator!(eval::SameInstancesEvaluator, generator::AbstractModelGenerator; seed=nothing)
-    instances = Array{CPModel}(undef, eval.nb_instances)
-    for i in 1:eval.nb_instances
+function SameInstancesEvaluator(valueSelectionArray::Array{H, 1}, generator::AbstractModelGenerator; seed=nothing, eval_freq = 50, nb_instances = 10) where H<: ValueSelection
+    instances = Array{CPModel}(undef, nb_instances)
+    metrics = Array{Any,2}(undef,nb_instances, size(valueSelectionArray,1))
```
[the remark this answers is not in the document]
3rdCore: In fact, it doesn't work because this kind of declaration doesn't allow the Array to contain basicmetrics with different signature.
As long as the first one is set, let's say basicmetrics{TakeObjective,learnedheuristic{SR, R, A}}, all the metrics in the array as to be of the same type.

PierreTsr (approved):

qcappart:
Good to me ! (provided that the last build passes :-)

I am very eager to see these plots on the next experiments on the learning !

qcappart (approved):

tomsander1998 (approved):

PR #142, opened 2021-05-28
Problem/nqueens

tomsander1998:
- small syntax modification on the AllDiff constraint
- datagenerator for nqueens
- nice plot for the tri partite graph


qcappart (approved):
Good for me. :-)

PierreTsr (changes requested):
Good to me, but I'd add a small test for the datagen, just to be sure the code doesn't crash in the future.

PierreTsr (approved):

PR #152, opened 2021-06-04
Feature/reward

3rdCore:
This is my proposition to simplify the use of struct of subtype `LearningPhase `
The `Phase `argument called by `valueselection `heuristic is no longer a element of type `...Phase <: LearningPhase` but simply the Type `...Phase` itself. That saves an allocation at each step of the search. 

I also fixed a little bug on the `plotRewardVariation` function.

### DIFF
--- a/src/CP/valueselection/lh_utils.jl
+++ b/src/CP/valueselection/lh_utils.jl
@@ -120,15 +120,6 @@ RLBase.is_terminated(env::unmaskedCPEnv) = env.terminal
 RLBase.state(env::unmaskedCPEnv) = env.state
 RLBase.ActionStyle(::unmaskedCPEnv) = MINIMAL_ACTION_SET
 
-"""
-    set_metrics!(PHASE::T, lh::LearnedHeuristic, model::CPModel, symbol::Union{Nothing, Symbol}, x::Union{Nothing, AbstractIntVar}) where T <: LearningPhase 
-
-Call set_metrics!(::SearchMetrics, ...) on env.search_metrics to simplify synthax.
-Could also add it to basicheuristic !
-"""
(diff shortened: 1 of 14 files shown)


src/CP/valueselection/rewards/defaultreward.jl
```diff
@@ -14,11 +14,22 @@ DefaultReward(model::CPModel) = DefaultReward(0)
...
+
+    elseif symbol == :FoundSolution
+        lh.reward.value -= 1
+
+    elseif symbol == :Feasible 
+        lh.reward.value -= 1
+
+    elseif symbol == :BackTracking
+        lh.reward.value -= 1
+    end
```
3rdCore, line 32: Finally we don't need to declare another function : `function set_reward!(::SolutionFoundPhase, ... ` as the information about the search (` :SolutionFound`, `:Feasible`, `:BackTracking `) is already contained in the argument `symbol`. 

We simply need to inform the user that he could implement a reward that can consider different symbols.

src/CP/valueselection/lh_utils.jl
```diff
@@ -120,15 +120,6 @@ RLBase.is_terminated(env::unmaskedCPEnv) = env.terminal
...
 
-"""
-    set_metrics!(PHASE::T, lh::LearnedHeuristic, model::CPModel, symbol::Union{Nothing, Symbol}, x::Union{Nothing, AbstractIntVar}) where T <: LearningPhase 
-
-Call set_metrics!(::SearchMetrics, ...) on env.search_metrics to simplify synthax.
-Could also add it to basicheuristic !
-"""
-function set_metrics!(PHASE::T, lh::LearnedHeuristic, model::CPModel, symbol::Union{Nothing, Symbol}, x::Union{Nothing, AbstractIntVar}) where T <: LearningPhase
-    set_metrics!(PHASE, lh.search_metrics, model, symbol, x::Union{Nothing, AbstractIntVar})
-end
```
3rdCore, line 131: this block of code was simply an interface between `LearnedHeuristic `and `search_metrics`

PR #158, opened 2021-06-14
SeaPearl v0.3.0: RL pipeline refactoring and GPU support

PierreTsr:
This is the first draft of the major revision of SeaPearl we are workning on with @3rdCore and @tomsander1998  for the last 2 weeks.

# What's new ?

The main changes are related to the ReinforcementLearning pipeline. Our goal was to get rid of the `ArrayBufferStateRepresentation` that we considered inefficient and too restrictive. We instead provided SeaPearl with a new kind of states: the `AbstractTrajectoryState` which could be understood as a bridge between the inner `AbstractStateRepresentation` and the RL agent. We provided by default a `DefaultTrajectoryState` which only consists in a `FeaturedGraph`  and the index of the variable we are branching on. This `DefaultTrajectoryState` is thus a lighter version of the inner state representation, with no restriction whatsoever.

We also added a `BatchedDefaultTrajectoryState`, which contains all the information of a batch of `DefaultTrajectoryState` but with data formatted for efficient parallel computation (i.e. as 3D arrays). The size of these arrays is determined dynamically when sending the batch to the target device and creating the `BatchedDefaultTrajectoryState` (though the functor function).

The novelties are:

- new struct `DefaultTrajectoryState`;
- new struct `BatchedDefaultTrajectoryState`;
- obligation to provide an `AbstractTrajectoryState` for the `StateRepresentation`;
- a `GeometricFlux.GraphConv` dispatch for `BatchedDefaultTrajectoryState` (impossible to do parallel computation with the existing code of GeometricFlux);
- revision of `FlexGNN` for the new input type;
- easy conversion of `CPLayerGraph` to adjacency matrix using `LigthGraphs`;
- new subtypes of `AbstractTrajectoryState` (Tabular and NonTabular) to specify the behavior of RL.jl;
- possibility to update the features at each step of the search;
- GPU support for the GNN computation;
- existing code refactoring;
- some CP bugs were found and corrected;

# What's next ?

This PR isn't ready for a merge yet. I'm posting this to let everybody see the new code and debate of it. There still important features missing, among which:

- support for the `TsptwStateRepresentation`;
- complete testset of all the new code;
- update of the Zoo and new results on the existing experiences;
- add new GNN model support for the new syntax (currently the new code isn't compatible with the GeometricFlux API, and only the simplest GraphConv is available: do we want the possibility to use all of their models despite some of them being way less efficient, or should we rewrite efficient code for the one we want to use ?)
- eventually change some choices of implementation?

Every comment/proposition is welcome on this PR, and do not hesitate if you have questions as this changes the RL code of SeaPearl a lot!


tomsander1998:
great summary ! I can start updating the SeaPearlZoo this afternoon.

tomsander1998 (approved):

test/CP/core/search/ilds.jl
```diff
@@ -284,11 +301,12 @@ using DataStructures
             ),
             trajectory = RL.CircularArraySLARTTrajectory(
                 capacity = 500,
-                state = Matrix{Float32} => (11, 17, 1),
+                state = SeaPearl.DefaultTrajectoryState[] => (),
```
3rdCore, line 304: Why should we use a 1-element `Vector{SeaPearl.DefaultTrajectoryState}` as element of the `CircularArrayBuffer ` instead of a `SeaPearl.DefaultTrajectoryState` ?

test/experiment/training.jl
```diff
@@ -95,78 +103,88 @@ end
...
     initial_params = deepcopy(params(learnedHeuristic.agent.policy.learner.approximator.model))
 
     metrics,evalmetrics = SeaPearl.train!(
         valueSelectionArray=learnedHeuristic, 
         generator=generator,
-        nbEpisodes=3,
+        nbEpisodes=10,
         strategy=SeaPearl.DFSearch,
-        variableHeuristic=SeaPearl.MinDomainVariableSelection{false}(),
+        variableHeuristic=SeaPearl.MinDomainVariableSelection{true}(),
```
3rdCore, line 185: objective function ( number of used color ) was not used by the variableHeuristic.

test/CP/core/search/ilds.jl
PierreTsr: It's an empty `Vector{SeaPearl.DefaultTrajectoryState}`, and it's because of RL.jl. RL.jl is used to get a matrix-like state, and thus the typing in of RL.jl relies on the function `eltype(state)`, which isn't currently defined for the `DefaultTrajectoryState`.

Another solution would be to define `eltype(DefaultTrajectoryState)`, but it would be confusing in other contexts.

src/RL/nn_structures/nn_structures.jl
```diff
@@ -9,18 +10,23 @@ abstract type NNStructure end
 
 Make NNStructure able to work with batches.
 """
-function (nn::NNStructure)(x::AbstractArray{Float32,3})
-    batch_size = size(x)[end]
+(nn::NNStructure)(ts::AbstractTrajectoryState) = throw(ErrorException("missing function (::$(typeof(nn)))(::$(typeof(ts)))."))
+function (nn::NNStructure)(x::AbstractVector{<:TabularTrajectoryState})
+    batch_size = size(x, 3)
     qval = [nn(x[:, :, i]) for i in 1:batch_size]
     hcat(qval...)
```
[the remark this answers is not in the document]
PierreTsr: I agree, it's never used in our code and it's unefficient. But it also is a very generic way to define the forward function for a batch of states, and I think it would work in almost all cases when no more specific function is defined.

src/CP/valueselection/learning/environment.jl
```diff
@@ -0,0 +1,43 @@
...
+RLBase.state(env::CPEnv) = env.state
+RLBase.ActionStyle(::CPEnv) = FULL_ACTION_SET
+RLBase.state_space(::SeaPearl.CPEnv, ::Observation{Any}, ::DefaultPlayer) = nothing
+
+struct unmaskedCPEnv{TS} <: AbstractCPEnv{TS}
+    reward::Float32
+    terminal::Bool
+    state::TS
+    actions_index::Array{Int64, 1}
+end
```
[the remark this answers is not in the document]
PierreTsr: I haven't tackled this question yet, as I figured it might be a bit out of topic of the gpu/optimization refactoring. But I agree we should discuss on the usefulness of this sutrcture.

test/RL/representation/default/defaulttrajectorystate.jl
```diff
@@ -0,0 +1,67 @@
...
+            GeometricFlux.FeaturedGraph(graph; nf=rand(3,6)),
+            2
+        )
+
+        batched1 = ts1 |> cpu
+        batched2 = ts2 |> cpu
+        @test isa(batched1, SeaPearl.BatchedDefaultTrajectoryState{Float32})
+
+        @test all(ts1.fg.graph .== batched1.fg.graph[:, :, 1])
+        @test all(abs.(ts1.fg.nf .- batched1.fg.nf[:,:,1]) .< 1e-5)
```
3rdCore, line 20: Where does the approximation comes from ?

PierreTsr: We are forced to work with `Float32` with the batched version (for GPU compatibility, which are very unefficient on `Float64`). But the iterative one can use any datatype, and by default Julia takes `Float64` thus inducing approximation errors.

PR #159, opened 2021-06-22
Fix/dependencies

PierreTsr:
This is a small PR to move a few extra functions to a new repository.

Some new features recently added to SeaPearl, brought a lot of new dependencies. These dependencies bring more complexity to the compatibility tree of SeaPearl, despite the fact that they aren't critical to the module. In order to avoid the dependencies of SeaPearl to become too messy, I suggest to create a new repo `SeaPearlExtras` with all the plotting & logging related functions. Thus a user could choose wether or not to load all these additional packages and it would be easier to maintain.


PierreTsr:
(I have the repo ready locally, I just need @qcappart approval to upload it 🙂 )

qcappart:
Yes I agree to have the plotting/non-critical dependencies in another repository.

qcappart (approved):

PR #160, opened 2021-06-22
fix CPModification update with View Variables

### LINKED ISSUE #55: CPModification not correctly updated in presence of views
In this example, the `CPModification` object does not get correctly updated:
```julia
trailer = SeaPearl.Trailer()
vars = SeaPearl.AbstractIntVar[]

x = SeaPearl.IntVar(2, 3, "x", trailer)
push!(vars, x)
y = SeaPearl.IntVar(5, 15, "y", trailer)
ax = SeaPearl.IntVarViewMul(x, 3, "3x")
push!(vars, ax)

minusY = SeaPearl.IntVarViewOpposite(y, "-y")
push!(vars, minusY)
constraint = SeaPearl.SumGreaterThan(vars, 5, trailer)

toPropagate = Set{SeaPearl.Constraint}()
prunedDomains = SeaPearl.CPModification()
SeaPearl.propagate!(constraint, toPropagate, prunedDomains)
```
After the `propagate!` call, we have
```julia
prunedDomains = Dict{String,Array{Int64,N} where N}(
   "3x" => [6],
   "-y" => [-8, -9, -10, -11, -12, -13, -14, -15]
)
```
but the value `2` was removed from `x.domain` and this modification is not visible.

PierreTsr:
As observed almost a year ago in #55 , `CPModification` isn't updated correctly with `IntVarView` and `BoolVarView`.

I propose a fix by providing a new definition of `addToPrunedDomains` for VariableViews. After creating/updating the entry corresponding to the view, it recursively call the function for the parent variable until a root variable is found.

Fixes #55.

### DIFF
--- a/src/CP/core/cpmodification.jl
+++ b/src/CP/core/cpmodification.jl
@@ -1,0 +1,87 @@
+const CPModification = Dict{String, Union{Array{Int}, Array{Bool}, SetModification}}
+
+"""
+    merge!(prunedDomains::CPModification, newPrunedDomains::CPModification)
+
+Merge `newPrunedDomains` into `prunedDomains`, concatenating the arrays if concerning the same variable.
+"""
+function merge!(prunedDomains::CPModification, newPrunedDomains::CPModification)
+    for k in keys(newPrunedDomains)
+        if haskey(prunedDomains, k)
+            prunedDomains[k] = vcat(prunedDomains[k], newPrunedDomains[k])
+        else
+            prunedDomains[k] = newPrunedDomains[k]
+        end
+    end
+end
+
+"""
+    addToPrunedDomains!(prunedDomains::CPModification, x::IntVar, pruned::Array{Int})
+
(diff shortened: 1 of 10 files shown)


src/CP/core/cpmodification.jl
```diff
@@ -0,0 +1,87 @@
...
+    # Update the ViewVariable entry
+    if haskey(prunedDomains, x.id)
+        prunedDomains[x.id] = vcat(prunedDomains[x.id], Array(pruned))
+    else
+        prunedDomains[x.id] = Array(pruned)
+    end
+
+    # Update the ParentVariable entry
+    parentPruned = parentValue.([x], pruned)
+    addToPrunedDomains!(prunedDomains, x.x, parentPruned)
```
3rdCore, line 63: Nice bug track !

3rdCore (approved):
Nice work !

qcappart:
Nice catch ! :-)

PR #172, opened 2021-07-02
Removing GPU usage from Test

PierreTsr:
I just saw that we recently introduced GPU usage during the tests by mistake.

It's not necessarly an issue (it helped me realize that CUDA wasn't working properly on the CIRRELT config) but I'm not sure if it should be the default behavior. If no CUDA installation is detected it won't try to use it though.

What are your thoughts about this ?


qcappart:
Good catch ! In my opinion, GPU s should be disabled by default for testing, because most of the work is the tree exploration + propagation, which do not leverage GPU abilities.

PR #174, opened 2021-07-05
AllDifferent optimization

PierreTsr:
After running some tests on Eternity 2, I observed that the search was suprisingly slow espacially the AllDiff constraint.

Thus I optimizied the consrtaint a bit (still not perfect though), which results in a x2 speedup for eternity2 without learning. This optimization uses ReversibleSparseBitSets (as in Compact Table) to store the state of the edges of the bipartite graph between 2 propagations, rather than an expensive dictionnary.
I also spotted a lack with the ViewVariable => when modifying a ViewVariable, the RootVariable corresponding values were not added to CPModification (and many tests were not designed to spot this). I spotted this as it caused the new AllDiff code to fail the NQueens test.

### DIFF
--- a/src/CP/core/cpmodification.jl
+++ b/src/CP/core/cpmodification.jl
@@ -15,6 +15,19 @@ function merge!(prunedDomains::CPModification, newPrunedDomains::CPModification)
     end
 end
 
+function updateChildren!(prunedDomains::CPModification, x::Union{IntVar, BoolVar}, pruned::Union{Array{Int}, Array{Bool}, BitArray})
+    for y in x.children
+        childrenPruned = childrenValue.([y], pruned)
+        if haskey(prunedDomains, y.id)
+            prunedDomains[y.id] = vcat(prunedDomains[y.id], Array(childrenPruned))
+        else
+            prunedDomains[y.id] = Array(childrenPruned)
+        end
+    end
+    return
+end
+
+
 """
     addToPrunedDomains!(prunedDomains::CPModification, x::IntVar, pruned::Array{Int})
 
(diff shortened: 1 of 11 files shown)


src/CP/core/cpmodification.jl
```diff
@@ -15,6 +15,19 @@ function merge!(prunedDomains::CPModification, newPrunedDomains::CPModification)
...
+    for y in x.children
+        childrenPruned = childrenValue.([y], pruned)
+        if haskey(prunedDomains, y.id)
+            prunedDomains[y.id] = vcat(prunedDomains[y.id], Array(childrenPruned))
+        else
+            prunedDomains[y.id] = Array(childrenPruned)
+        end
+    end
+    return
+end
```
3rdCore, line 28: Nice spot ! How we couldn't think of this previously, this looks so obvious now ...

src/CP/variables/IntVarView.jl
```diff
@@ -264,4 +273,8 @@ end
 
 parentValue(y::IntVarViewMul, v::Int) = v ÷ y.a
 parentValue(y::IntVarViewOffset, v::Int) = v - y.c
-parentValue(::IntVarViewOpposite, v::Int) = - v
\ No newline at end of file
+parentValue(::IntVarViewOpposite, v::Int) = - v
+childrenValue(::IntVar, v::Int) = v
+childrenValue(y::IntVarViewMul, v::Int) = childrenValue(y.x, v) * y.a
```
3rdCore, line 278: Just to be sure, v represents the pruned value of the parent ?

3rdCore:
Looks good for me, nice work !

PR #180, opened 2021-07-13
Add an input layer for node and edge features

PierreTsr:
As discussed with @tomsander1998 and @3rdCore, we thought it could be smart to add node-wise layers before the graph layers. It would makes it possible to use normalization on input features, and efficiently resize the features recieved by the GNN.
This has not been tested in depth yet, and it might still be a bit early to merge it.


3rdCore:
This PR has been discontinued. Indeed, CPNN and FullFeaturedCPNN will get a huge refacto to handle heterogeneous convolution.

3rdCore:
This PR has been discontinued. Indeed, CPNN and FullFeaturedCPNN will get a huge refacto to handle heterogeneous convolution.

PR #185, opened 2021-07-26
Timeout feature on evaluation 

3rdCore:
Few modifications on CPModel to add the timeout feature on searches. CPModel now has a `limit.searchingTime` attribute. 
To use a timeout on evaluation, the user should add the optional argument `evalTimeOut` during the `SameInstancesEvaluator` creation. 

I added the possibility to define 2 different strategies `strategy` and `eval_strategy` for both the training and the evaluation. 

I fixed a little omission on `BasicHeuristic` that were previously not modifying `model.statistics.lastVar` during a _DecisionPhase_

Prints during search, using `verbose=true` have also been updated.


src/CP/core/model.jl
```diff
@@ -192,10 +193,10 @@ end
 
 Check if `model`' statistics are still under the limits.
 """
-belowLimits(model::CPModel) = belowNodeLimit(model) && belowSolutionLimit(model)
+belowLimits(model::CPModel) = belowNodeLimit(model) && belowSolutionLimit(model) && belowTimeLimit(model)
 belowNodeLimit(model::CPModel) = isnothing(model.limit.numberOfNodes) || model.statistics.numberOfNodes < model.limit.numberOfNodes
 belowSolutionLimit(model::CPModel) = isnothing(model.limit.numberOfSolutions) || model.statistics.numberOfSolutions < model.limit.numberOfSolutions
-
+belowTimeLimit(model::CPModel) = isnothing(model.limit.searchingTime) || peektimer() < model.limit.searchingTime
```
3rdCore, line 199: `peektimer()` can be compared as the loop button on any stopwatch. It gives the elapsed time since the last call to `tic()`.

src/CP/core/search/search.jl
```diff
@@ -19,7 +19,7 @@ at each branching and using `valueSelection` to choose how the branching will be
 
 """
 function search!(model::CPModel, strategy::S, variableHeuristic::AbstractVariableSelection, valueSelection::ValueSelection=BasicHeuristic(); out_solver::Bool=false) where S <: SearchStrategy
-
+    tic()
```
3rdCore, line 22: Timer is started at the beginning of a search.

PierreTsr (approved):
Everything's good to me

PR #186, opened 2021-07-28
Replacing GeometricFlux.jl to boost performances

PierreTsr:
WiP


src/RL/utils/featuredgraph/featuredgraph.jl
```diff
@@ -0,0 +1,90 @@
+mutable struct FeaturedGraph{T <: AbstractMatrix, N <: AbstractMatrix, E <: AbstractArray, G <: AbstractVector} <: AbstractFeaturedGraph
+    graph::T
+    nf::N
+    ef::E
+    gf::G
+    directed::Bool
```
3rdCore, line 6: What is the benefit of using parametric types `{T <: AbstractMatrix, N <: AbstractMatrix, E <: AbstractArray, G <: AbstractVector}` for the `FeaturedGraph` struct ? Is it from GeometricFlux ?

3rdCore (commented):
still reviewing

src/RL/utils/featuredgraph/batchedfeaturedgraph.jl
```diff
@@ -0,0 +1,114 @@
...
+    gf = zeros(T, gfLength, ngraphs)
+
+    for (i, fg) in enumerate(fgs)
+        graph[1:nv(fg),1:nv(fg),i] = fg.graph
+        nf[:, 1:nv(fg), i] = fg.nf
+        ef[:, 1:nv(fg), 1:nv(fg), i] = fg.ef
+        gf[:, i] = fg.gf
+    end
+
+    return BatchedFeaturedGraph{T}(graph, nf, ef, gf)
```
3rdCore, line 52: Looks good to me !

src/RL/utils/featuredgraph/batchedfeaturedgraph.jl
```diff
@@ -0,0 +1,114 @@
...
+end
+
+BatchedFeaturedGraph(fgs::Vector{FG}) where {FG <: FeaturedGraph} = BatchedFeaturedGraph{Float32}(fgs)
+
+# ========== Accessing ==========
+# code from GraphSignals.jl
+
```
3rdCore, line 59: Is the following code used elsewhere in the project ?

src/RL/utils/featuredgraph/featuredgraph.jl
3rdCore: Yes, it's from GeometricFlux.

PierreTsr:
The PR is complete and all test were working fine. But there seems to be an issue with one of our dependencies, that appeared only recently. I won't be able to investigate further this issue, but I don't think it's wise to merge a PR with a failing test. I don't think the problem is related with the new code, only with the new modules versions that were previously restricted by GeometricFlux.

Outside of this, SeaPearl is working just fine and I have been able to run many experiments with this code.

CarloLucibello:
Hi people, 
just a heads up: 
I'm registering a new GNN package
https://github.com/CarloLucibello/GraphNeuralNetworks.jl
which hopefully fixes the issues you experienced with GeometricFlux 
(it has batched graphs, good cuda support, sparse matrix operations when possible, much faster and scalable message passing, etc...).
I see you have done a ton of work here, but maybe you don't want the burden of maintaining a full-fledged GNN library.
If you agree I could cook a PR with GraphNeuralNetworks.jl support and/or fix there any additional requirements you have for SeaPearl

PR #187, opened 2021-08-05
Fixing major leak between training and evaluation data

PierreTsr:
After a close inspection we detected with @3rdCore an issue in the evaluation pipeline.

Until now SeaPearl has been relying on the function `Flux.testmode!` to prevent training of the NN model on test samples. However this function doesn't prevent the addition of these samples to the memory of the RL agent. We beleive that these samples are then used by RL.jl sampler to provide training samples to the agent. This behavior could explain some other strange behaviors that were observed in the last months (decreasing reward on the train samples, learning despite erroneous reward...).

The fix to this issue is rather simple, and consists in preventing RL.jl to do any update when the heuristic is set in test mode.

### DIFF
--- a/src/RL/representation/tsptw/tsptwstaterepresentation.jl
+++ b/src/RL/representation/tsptw/tsptwstaterepresentation.jl
@@ -29,13 +29,14 @@ end
 
 TsptwStateRepresentation(model::CPModel) = TsptwStateRepresentation{TsptwFeaturization, TsptwTrajectoryState}(model)
 
+# This function is used for legacy. It enables compatibility with the experiments present during CPAIOR 2021, with variableOutputGNN
 function TsptwTrajectoryState(sr::TsptwStateRepresentation{F, TsptwTrajectoryState}) where F
     # TODO change this once the InitializingPhase is fixed
     if isnothing(sr.variableIdx)
         sr.variableIdx = 1
     end
     if isnothing(sr.possibleValuesIdx)
-        throw(ErrorException("Unable to build a DefaultTrajectoryState, when the possible values vector is nothing."))
+        throw(ErrorException("Unable to build a TsptwTrajectoryState, when the possible values vector is nothing."))
     end
 
     n = size(sr.dist, 1)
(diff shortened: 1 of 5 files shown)


qcappart (approved):
Great work for spotting this important issue.

Tiny nitpicking: trainmode -> trainMode

src/RL/representation/tsptw/tsptwstaterepresentation.jl
```diff
@@ -29,13 +29,14 @@ end
...
 
+# This function is used for legacy. It enables compatibility with the experiments present during CPAIOR 2021, with variableOutputGNN
 function TsptwTrajectoryState(sr::TsptwStateRepresentation{F, TsptwTrajectoryState}) where F
     # TODO change this once the InitializingPhase is fixed
     if isnothing(sr.variableIdx)
         sr.variableIdx = 1
     end
     if isnothing(sr.possibleValuesIdx)
-        throw(ErrorException("Unable to build a DefaultTrajectoryState, when the possible values vector is nothing."))
+        throw(ErrorException("Unable to build a TsptwTrajectoryState, when the possible values vector is nothing."))
```
3rdCore, line 39: 👍

src/RL/representation/tsptw/tsptwstaterepresentation.jl
```diff
@@ -45,6 +46,25 @@ function TsptwTrajectoryState(sr::TsptwStateRepresentation{F, TsptwTrajectorySta
...
+    n = size(sr.dist, 1)
+    adj = ones(Int, n, n) - I
+    edgeFeatures = build_edge_feature(adj, sr.dist)
+    fg = GeometricFlux.FeaturedGraph(adj; nf=sr.nodeFeatures, ef=edgeFeatures)
+
+    actionSpace = collect(1:n)
+
+    return DefaultTrajectoryState(fg, sr.variableIdx, actionSpace)
+end
+
```
3rdCore, line 67: Is this an artifact from a previous PR ?

src/experiment/evaluation.jl
```diff
@@ -24,19 +24,24 @@ function SameInstancesEvaluator(valueSelectionArray::Array{H, 1}, generator::Abs
...
+            dt = @elapsed search!(model, strategy, variableHeuristic, heuristic)
             eval.metrics[i,j](model,dt)
 
-            verbose && println(typeof(eval.metrics[1,j].heuristic), " evaluated with: ", model.statistics.numberOfNodes, " nodes, taken ", dt, "s, number of solutions found : ", model.statistics.numberOfSolutions)
+            verbose && println(typeof(heuristic), " evaluated with: ", model.statistics.numberOfNodes, " nodes, taken ", dt, "s, number of solutions found : ", model.statistics.numberOfSolutions)
         end 
-        testmode!(eval.metrics[1,j].heuristic, false)
+        testmode!(heuristic, false)
+        if !isnothing(initsize)
+            @assert length(heuristic.agent.trajectory) == initsize "You have leaks in your evaluation pipeline!"
```
3rdCore, line 43: Nice warning !

3rdCore (changes requested):
Nice work !
current tests should not be impacted by this modification unless some of them manually create a learnedHeuristic without using the implemented constructor : ` LearnedHeuristic{SR, R, A}(agent::RL.Agent) `. 

Before merging, I would be nice to have some testset that validates this new features, I can write them if you want.

src/RL/representation/tsptw/tsptwstaterepresentation.jl
PierreTsr: nope, it's a missing function to enable DefaultTrajectoryState to work with TSPTW

PierreTsr:
The PR is complete and everything is working fine.

However, we recently found with @3rdCore that one of our dependcies is causing some test completely unrelated with our changes to fail. We haven't been able to find where this issue is coming form. It is very strange because on my side I have been able to conduct many experiments without any issue. I don't know if the PR should be merged or not, as the fails aren't related with the new code.

PierreTsr:
I tried to add a testset, but the recent issue we encountered with @3rdCore, likely caused by one of our dependency and not related to the above code, makes it impossible to check if the test is correctly executed.

However I have been able to use this code to run a few experiments so everything else should be fine.

PR #189, opened 2021-10-19
Variant of CPReward

sandertom:
I called the variant of CPReward with a second order term CPReward2. It can be interesting to compare both. There is no other modification.


3rdCore:
Interesting CPreward, I approve the PR but some testsets would be nice to have for that reward in the mid-term...

PR #192, opened 2022-01-18
fix type in feature_length

marco-novaes98:
First pull request by Marco as test, just to fix a minor bug

### DIFF
--- a/src/RL/representation/default/defaultstaterepresentation.jl
+++ b/src/RL/representation/default/defaultstaterepresentation.jl
@@ -81,6 +81,6 @@ end
 
 Returns the length of the feature vector, for the `DefaultFeaturization`.
 """
-feature_length(::Type{<:FeaturizedStateRepresentation{DefaultFeaturization, TS}}) where TS <: DefaultTrajectoryState = 3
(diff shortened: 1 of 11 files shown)


3rdCore:
your latest commit didn't pass all test sets. This was due to some Core modifications of the direct dependencies _ReinforcementLearning.jl_. 

The random number generator also changed a bit between Julia v1.6 and v1.7, Ialso  added new test sets to handle both cases.

src/RL/representation/default/defaultstaterepresentation.jl
```diff
@@ -81,6 +81,6 @@ end
 
 Returns the length of the feature vector, for the `DefaultFeaturization`.
 """
-feature_length(::Type{<:FeaturizedStateRepresentation{DefaultFeaturization, TS}}) where TS <: DefaultTrajectoryState = 3
+feature_length(::Type{<:FeaturizedStateRepresentation{DefaultFeaturization, TS}}) where TS = 3
```
3rdCore, line 84: I finally removed the rapid fix that we proposed yesterday. There was initially no reason to specify the parametric type `TS`, the error we encountered before could have been the result of an uncanny artifact due to conflict between updated dependencies and our older Julia v1.6...

marco-novaes98:
All problems have been fixed.

PR #203, opened 2022-02-23
Added features for experiment reproductibility and many other improvement : 

3rdCore:
I added several interesting features to be able to exactly reproduce each experiment given a couple of seed `(seedEval,seed)` : 

- `seedEval` is used to generate a specific random stream to generate evaluation instances. 
- `seed` is used to handle every other random stream. (random heuristic, weight initialisation...)

I plotted some metrics for 2 different experiments on a _tsptw_ experiment with 20 nodes. It is absolutely normal that we see no improvement of the learned heuristics as hypermeters were not set for this purpose.  As we can see performance of the nearest heuristic is the same in both experiments, showing good proof that Evaluation instances are consistent given the same `Evalseed` : 

Experiment 1             |  Experiment 2
:-------------------------:|:-------------------------:
image | image

This work is still in progress, as I am facing some unexpected behaviors from generated CP instances.  Given an exact same model, the fix-point yields different results in terms of value order for a given variable. As a result, random heuristic choices are not consistent across experiments as shown in previous plots. The ollowing screenshots shows that 2 identical assignation yield 2 different domain indexing. 

Experiment 1             |  Experiment 2
:-------------------------:|:-------------------------:
image| image

Features added : 

- [x] added seed for evaluation instance generation
- [x] added seed for random ValueHeuristic
- [x] added seed for RL explorer

Minor changes/bug fixed : 

- [x] evalFreq can no longer be smaller than 1, (in that case they were no evaluations)
- [x] An evaluation is now done before running a first training episode
- [x] added AccumulatedRewardBeforeReset to retrieve one episode total reward
- [x] determinist and random heuristics are no longer re-evaluated along with training as their behavior is not supposed to change. 
- [x] SmartReward has been totally modified. It is now the exact same one used by @qcappart in its 2020 paper. 
- [x] Added `TsptwGeneratorFromRealData` generator based on @kimriouxparadis work. 

What will come next / things to fix : 

- [x] During the first evaluation, execution time takes into account precompiling time which can biased the metric. that need to be fixed.
- [x] testset for every new features
- [x] add seed for RL model weight initialization
- [x] add seed for training instance generation
- [x] refacto all `fillwithgenerator ` functions for every generator
- [ ] make sure that `TsptwGeneratorFromRealData` works with SmartReward which is not the case by now as `TsptwGeneratorFromRealData` does not fill `model.adhocinfo` with the same attributes than 'TsptwGenerator' --> POSTPONED


src/CP/valueselection/learning/rewards/smartreward.jl
```diff
@@ -19,15 +23,19 @@ function set_reward!(::Type{StepPhase}, lh::LearnedHeuristic{SR, SmartReward, A}
...
+        n =  size(dist)[1]
+        max_dist = Float32(Base.maximum(dist))
+            if isbound(model.variables["a_"*string(n-1)])
+                last = assignedValue(model.variables["a_"*string(n-1)])
+                first = assignedValue(model.variables["a_1"])
+        
+                dist_to_first_node = lh.current_state.dist[last, first] * max_dist
+                print("final_dist : ", dist_to_first_node, " // ")
+                lh.reward.value += -ρ*dist_to_first_node 
+            end
```
3rdCore, line 37: This corresponds to the last path portion between the last assigned variable and the start variable.

src/CP/valueselection/learning/rewards/smartreward.jl
```diff
@@ -38,16 +46,34 @@ end
...
+    max_dist = Float32(Base.maximum(dist))
+
+    if !isnothing(model.statistics.lastVar)
+        x = model.statistics.lastVar
+        s = x.id
+        current = parse(Int,split(x.id,'_')[2])
+        if isbound(model.variables["a_"*string(current)])
+            a_i = assignedValue(model.variables["a_"*string(current)])
+            v_i = assignedValue(model.variables["v_"*string(current)])
+            last_dist = lh.current_state.dist[v_i, a_i] * max_dist
```
3rdCore, line 69: This corresponds to the distance between the previous node and the node that has just been selected by the heuristic one step before. (Recall that the reward is always given one step after, just before making a new decision).

src/CP/valueselection/learning/utils.jl
```diff
@@ -9,6 +9,7 @@ manually change the mode again if he wants.
 """
 function Flux.testmode!(lh::LearnedHeuristic, mode = true)
     Flux.testmode!(lh.agent, mode)
+    lh.agent.policy.explorer.is_training = !mode
```
3rdCore, line 12: fixing RL agent's explorer value to zero during evaluation.

src/RL/utils/totalreward.jl
```diff
@@ -1,13 +1,15 @@
...
     last_episode_total_reward(t::AbstractTrajectory)
 
-Compute the sum of every reward of the last episode of the trajectory
+Compute the sum of every reward of the last episode of the trajectory. 
+
+For example, if the t[:terminal] = [0, 0, 1, 0, 1, 1, 1, 0, 0, 1], The 7-th state is a terminal state, which means that the last episode started at step 8. Hence, last_episode_total_reward corresponds to the 3 lasts decisions.
 """
 function last_episode_total_reward(t::AbstractTrajectory)
     last_index = length(t[:terminal])
+    last_index == 0 && return 0
```
3rdCore, line 10: return 0 in case the trajectory is empty. This was needed in order to evaluate the model before any training step without triggering a `BoundsError` while retrieving `last_episode_total_reward` in the trajectory.

src/datagen/tsptw.jl
```diff
@@ -21,20 +21,21 @@ https://arxiv.org/abs/2006.01610
 
 Basicaly finds positions with a uniform distributions, then sets the time windows by creating a feasible tour and adding
 some randomness by using uniform distributions with gap and the length of the time windows.
+
+A seed must be specified by the user to generate a specific instance. As long as Random.seed!(seed) is called at the beginning of the function, every random-based operations with be deterministic. Caution : this is not the seed that must be specified in order to generate a same set of evaluation instances across experiment.
 """
-function fill_with_generator!(cpmodel::CPModel, gen::TsptwGenerator; seed=nothing, dist = nothing, timeWindows=nothing)
-    if !isnothing(seed)
-        Random.seed!(seed)
-    end
```
3rdCore, line 28: We want evaluation instances to be the same across experiments not to be the same inside an experiment (which is totally the opposite we are trying to achieve), hence the `rng` attribute has to be defined outside the `fill_with_generator!` function.

src/experiment/evaluation.jl
```diff
@@ -8,40 +23,67 @@ mutable struct SameInstancesEvaluator <: AbstractEvaluator
...
             @assert length(heuristic.agent.trajectory) == initsize "You have leaks in your evaluation pipeline!"
+            
+        
+        else                                            #BasicHeuristic only needs to be evaluated once, as it is not updated during training. As long as the eval has already been done once, it only repeat the search metrics for each evaluation step; 
+            for i in 1:eval.nbInstances
+                model = eval.instances[i]
+                reset_model!(model)
+                if eval.metrics[i,j].nbEpisodes == 0
+                dt = @elapsed search!(model, strategy, variableHeuristic, heuristic)
+                eval.metrics[i,j](model,dt)
```
3rdCore, line 77: A search is computed for BasicHeuristics only once.

src/experiment/evaluation.jl
```diff
@@ -8,40 +23,67 @@ mutable struct SameInstancesEvaluator <: AbstractEvaluator
...
+        else                                            #BasicHeuristic only needs to be evaluated once, as it is not updated during training. As long as the eval has already been done once, it only repeat the search metrics for each evaluation step; 
+            for i in 1:eval.nbInstances
+                model = eval.instances[i]
+                reset_model!(model)
+                if eval.metrics[i,j].nbEpisodes == 0
+                dt = @elapsed search!(model, strategy, variableHeuristic, heuristic)
+                eval.metrics[i,j](model,dt)            
+                verbose && println(typeof(heuristic), " evaluated with: ", model.statistics.numberOfNodes, " nodes, taken ", dt, "s, number of solutions found : ", model.statistics.numberOfSolutions)
+                else
+                dt,numberOfNodes, numberOfSolutions = repeatlast!(eval.metrics[i,j])
```
3rdCore, line 80: For every following evaluation step, we only repeat the metrics and avoid computing already existing results.

3rdCore:
NN models can now completely be deterministically generated. One step toward perfect experiment reproductibility.

3rdCore:
The first evaluation is not anymore impacted by compilation time required while running the `evaluate` code block for the first time.

src/datagen/tsptwRealData.jl
```diff
@@ -0,0 +1,215 @@
...
+        timeWindows[i, 2] = maxValue
+    end
+
+    random_solution = [1, shuffle(rng, Vector(2:gen.n_city))]
+
+    #TODO need to find a way to retrieve coordinates of points from 
+    #x_pos = zeros(gen.n_city)
+    #y_pos = zeros(gen.n_city)
+    #grid_size = 0
+    #cpmodel.adhocInfo = dist, timeWindows, hcat(x_pos, y_pos), grid_size
```
gostreap, line 86: It seems to me that in order to use the tsptw specific rewards, you have to fill in `adhocInfo`.

Since this new generator does not seem essential to me, maybe just remove it from the pull request until it is functional.

gostreap (approved):
For me there is no major problem to merge this pull request.

Maybe we should just postpone the addition of `tsptwRealData.jl`, since it doesn't seem essential and it doesn't seem to me to be fully functional.

Apart from that, there are plenty of useful improvements and the only remarks are about comments or minor points.

src/datagen/tsptwRealData.jl
3rdCore: That's right, I moved the file `tsptwRealData.jl` and its dependencies on the branch [Tsptw-Real-Data](https://github.com/corail-research/SeaPearl.jl/tree/Tsptw-Real-Data).

src/datagen/latin.jl
```diff
@@ -30,8 +22,18 @@ function proper_move(A::Matrix{Int},x::Int,y::Int,z::Int,v::Int)
     end
 end
 
+"""
+    fill_with_generator!(cpmodel::CPModel, gen::GraphColoringGenerator)
+Fill a CPModel with the variables and constraints generated. We fill it directly instead of
+creating temporary files for efficiency purpose.
+
+A seed must be specified by the user to generate a specific instance. As long as Random.seed!(seed) is called at the beginning of the function, every random-based operations with be deterministic. Caution : this is not the seed that must be specified in order to generate a same set of evaluation instances across experiment, in that case, the user must call Random.seed! only once, at the beginning of the experiment.
```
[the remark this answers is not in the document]
3rdCore: > It seems to me that by default the explorer is not called when `trainMode` is false.

I thought that it was the case too, but it is not. Before this modifications, the explorer rate was decreasing even during evaluation which was highly problematic.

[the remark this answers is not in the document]
3rdCore: We discussed of this with the creator of the package ReinforcementLearning.jl [here.](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/discussions/585)

src/datagen/nqueens.jl
```diff
@@ -15,14 +15,13 @@ end
 Fill a CPModel with the variables and constraints generated. We fill it directly instead of
 creating temporary files for efficiency purpose.
 
+A seed must be specified by the user to generate a specific instance. As long as Random.seed!(seed) is called at the beginning of the function, every random-based operations with be deterministic. Caution : this is not the seed that must be specified in order to generate a same set of evaluation instances across experiment, in that case, the user must call Random.seed! only once, at the beginning of the experiment. 
+
 This generator create graps for the NQueens problem.
 
 """
-function fill_with_generator!(cpmodel::CPModel, gen::NQueensGenerator; seed=nothing)
+function fill_with_generator!(cpmodel::CPModel, gen::NQueensGenerator; rng::Union{Nothing,AbstractRNG} = nothing)
```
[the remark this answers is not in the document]
3rdCore: That's right, I removed it.

src/experiment/metrics/basicmetrics.jl
```diff
@@ -91,37 +86,24 @@ function (metrics::BasicMetrics{TakeObjective, <:LearnedHeuristic})(model::CPMod
...
-function computemean!(metrics::BasicMetrics{O, H}) where {O, H<:ValueSelection} 
-    windowspan = min(metrics.meanOver,length(metrics.meanNodeVisitedUntilfirstSolFound))
-    nodeVisitedUntilFirstSolFound = copy(metrics.meanNodeVisitedUntilfirstSolFound)
-    metrics.meanNodeVisitedUntilfirstSolFound = rollmean(nodeVisitedUntilFirstSolFound, windowspan)
-    nodeVisitedUntilOptimality = copy(metrics.meanNodeVisitedUntilEnd)
-    metrics.meanNodeVisitedUntilEnd = rollmean(nodeVisitedUntilOptimality, windowspan)
-
-    if isa(metrics,BasicMetrics{O,<:LearnedHeuristic})
-        metrics.totalReward = rollmean(metrics.totalReward,windowspan)
+function repeatlast!(metrics::BasicMetrics{<:AbstractTakeObjective, <:BasicHeuristic})
```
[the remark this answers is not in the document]
3rdCore: This function is really confusing but was an easy way to virtually repeat evaluation metrics without having to do 1 week of refacto.

marco-novaes98 (commented):
This is not an exhaustive review, there are a lot of changes and I don't understand all the details of the RL part but I tried to roughly understand the changes and check if it made sense to me.

src/RL/utils/totalreward.jl
```diff
@@ -1,13 +1,15 @@
...
 
-    #if t[:terminal][last_index]   #TODO understand why they wrote this
-
+    #if t[:terminal][last_index]  Do we need to consider cases where the last state is not a terminal state ?
```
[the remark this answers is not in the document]
3rdCore: Yes @marco-novaes98, this case is done in line 10.

src/CP/valueselection/learning/rewards/smartreward.jl
```diff
@@ -19,15 +23,18 @@ function set_reward!(::Type{StepPhase}, lh::LearnedHeuristic{SR, SmartReward, A}
...
-
+    elseif symbol == :FoundSolution #last portion required to get the full closed path
+        dist = model.adhocInfo[1]
+        n =  size(dist)[1]
+        max_dist = Float32(Base.maximum(dist))
+            if isbound(model.variables["a_"*string(n-1)])
+                last = assignedValue(model.variables["a_"*string(n-1)])
+                first = assignedValue(model.variables["a_1"])
+        
+                dist_to_first_node = lh.current_state.dist[last, first] * max_dist
```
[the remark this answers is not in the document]
3rdCore: This is supposed to be copied from the original implementation of the reward provided by @qcappart : 

https://github.com/qcappart/hybrid-cp-rl-solver/blob/master/src/problem/tsptw/environment/environment.py

However as I can't find it anymore in the original repo, I removed the factor `* max_dist`  in the computation of `last_dist` and `dist_to_first_node`.

src/experiment/metrics/basicmetrics.jl
```diff
@@ -52,9 +51,8 @@ function (metrics::BasicMetrics{DontTakeObjective, <:BasicHeuristic})(model::CPM
     metrics.nbEpisodes+=1
     push!(metrics.nodeVisited,copy(model.statistics.nodevisitedpersolution))
     push!(metrics.meanNodeVisitedUntilEnd,model.statistics.numberOfNodes)
-    if ! isempty(model.statistics.nodevisitedpersolution)    #infeasible case
-        push!(metrics.meanNodeVisitedUntilfirstSolFound,model.statistics.nodevisitedpersolution[1])
-    end    
+    index = findall(!isnothing, model.statistics.solutions)  #return the list of index of real solution in model.statistics.solutions
+    push!(metrics.meanNodeVisitedUntilfirstSolFound, !isempty(index) ? model.statistics.nodevisitedpersolution[index[1]] : nothing)
```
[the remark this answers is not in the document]
3rdCore: In the general case, we can have more than one solution. 

> If so, I don't understand why we only consider the first solution in the push! right after.
This is because we are looking for the number of nodes until the first solution is found : 
`    push!(metrics.meanNodeVisitedUntilfirstSolFound, !isempty(index) ? model.statistics.nodevisitedpersolution[index[1]] : nothing)
`

> If not, it might be better to use findfirst to make the code clearer and a bit faster (and also update the comment)
Your point is correct, `findfirst` will be the best practice.

3rdCore:
Fix some issues raised by @marco-novaes98. 

This PR can finally be merged.

PR #205, opened 2022-03-16
Fixed testsets due to updates on ReinforcementLearning.jl package. 

3rdCore:
_Reminder_ : For RL-based agent, SeaPearl stores the trajectory in a object called `CircularArraySARTTrajectory`, which contains several CircularArrayBuffer to store each sub-class(`state`, `action`, `reward`, and `terminal`) for successive training steps.

Due to changes on [CircularArrayBuffer.jl](https://github.com/JuliaReinforcementLearning/CircularArrayBuffers.jl) which is a direct dependency of ReinforcementLearning.jl, `CircularArrayBuffer` object expects samples to be _iterable_ (in order to be able to append more than one sample all at once, Which is a useful feature for Array-like objects but that is irrelevant for single-element object. 

However, for unknown reasons I had to deactivate some testsets regarding the RBSearch strategy when being coupled with PPO-based RL agent. I'd be nice to figure out why these tests don't pass anymore, but the stackTrace refers to unusual functions of the package _CircularArrayBuffer.jl_, and digging into the problem would certainly take hours...

### DIFF
--- a/src/CP/valueselection/learning/learnedheuristic.jl
+++ b/src/CP/valueselection/learning/learnedheuristic.jl
@@ -24,6 +24,22 @@ learns thanks to rewards that are given regularly during the search. He wil try 
 From the RL point of view, this LearnedHeuristic also plays part of the role normally played by the environment. Indeed, the
 LearnedHeuristic stores the action space, the current state and reward. The other role that a classic RL environment has is describing
 the consequences of an action: in SeaPearl, this is done by the CP part - branching, running fixPoint!, backtracking, etc...
+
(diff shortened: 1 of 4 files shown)


3rdCore:
Bugs on testsets were not related to SeaPearl internal process. In fact, it was an edge-case on the CircularArrayBuffer.jl package identified [here](https://github.com/JuliaReinforcementLearning/CircularArrayBuffers.jl/issues/12).

3rdCore:
The issues have been fixed on SeaPearl dependencies.

PR #207, opened 2022-03-31
Add summary of the performance test 

marco-novaes98:
Summary of the performance test  between Gecode and SeaPearl. The objective is to compare the execution time between SeaPearl and a commercial CP solver.

### DIFF
--- a/docs/src/CP/performance_test.md
+++ b/docs/src/CP/performance_test.md
@@ -18,7 +18,7 @@ A basic model has been implemented in MiniZinc and SeaPearl using the same logic
 
 The main observations from the experience are:
 
-* The time of execution grows over-polynomially with the size of the instance. This result was expected as the KEP is a combinatorial problem.
+* The time of execution grows over-polynomially with the size of the instance. This result was expected for this combinatorial problem.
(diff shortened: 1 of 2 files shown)


docs/src/CP/performance_test.md
```diff
@@ -0,0 +1,31 @@
...
+
+The problem used for the experience is the Kidney Exchange Problem (KEP) using 7 instances from size 5 to 35 (step size of 5). 
+
+A basic model has been implemented in MiniZinc and SeaPearl using the same logic (variables, constraints, heuristics...).
+
+## Results and conclusions
+
+The main observations from the experience are:
+
+* The time of execution grows over-polynomially with the size of the instance. This result was expected as the KEP is a combinatorial problem.
```
3rdCore, line 21: In fact, the over-polynomial behavior is not specifically a characteristic of combinatorial problems even if in practice, most of the problems suffers from a exponential time complexity for solving w.r.t input size.

marco-novaes98: All right, I have modified this sentence to be more precise

PR #208, opened 2022-04-05
Override the - operator to easily define constraints of the form y= -x. 

### LINKED ISSUE #198: [CP] No easy way to generate the opposite of a variable, y = -x
By now, to generate the opposite of a variable `y = -x`, one needs to proceed as follow : 

```julia
x = SeaPearl.IntVar(1, 5, "x", trailer)
y = SeaPearl.IntDomainViewOpposite(x.domain)
```

It would be nice to somehow overwrite the base operator `-` to automatically generate a `IntDomainViewOpposite` variable. This would be a good feature to facilitate future CP problem implementation and can be a nice good first issue for a newcomer.

gostreap:
Closes #198.

The tests performed are quite simple, should we go further?

### DIFF
--- a/test/CP/variables/IntVar.jl
+++ b/test/CP/variables/IntVar.jl
@@ -45,12 +45,15 @@
         x = SeaPearl.IntVar(5, 8, "x", trailer)
         y = -x
         
-        @test isa(y, SeaPearl.IntDomainViewOpposite)
-        @test SeaPearl.minimum(y) == -8
-        @test SeaPearl.maximum(y) == -5
+        @test isa(y, SeaPearl.IntVarViewOpposite)
+        @test SeaPearl.minimum(y.domain) == -8
+        @test SeaPearl.maximum(y.domain) == -5
+        @test !SeaPearl.isbound(y)
 
         SeaPearl.assign!(x, 5)
 
-        @test SeaPearl.length(y) == 1
+        @test SeaPearl.length(y.domain) == 1
+        @test SeaPearl.isbound(y)
+        @test SeaPearl.assignedValue(y) == -5
     end
 end
--- a/test/CP/variables/IntVar.jl
+++ b/test/CP/variables/IntVar.jl
@@ -39,4 +39,18 @@
 
         @test SeaPearl.assignedValue(x) == 5
     end
+
+    @testset "-()" begin
+        trailer = SeaPearl.Trailer()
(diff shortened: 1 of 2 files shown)


louis-gautier (approved):

test/CP/variables/IntVar.jl
```diff
@@ -39,4 +39,18 @@
...
+        x = SeaPearl.IntVar(5, 8, "x", trailer)
+        y = -x
+        
+        @test isa(y, SeaPearl.IntDomainViewOpposite)
+        @test SeaPearl.minimum(y) == -8
+        @test SeaPearl.maximum(y) == -5
+
+        SeaPearl.assign!(x, 5)
+
+        @test SeaPearl.length(y) == 1
```
3rdCore, line 54: Nice tests. I'd just add a test on the value of `y` after assigning x.

3rdCore (approved):

test/CP/variables/IntVar.jl
gostreap: Actually, y is a view on the domain, so I don't think there is a way to test the value, but maybe it would be more logical to return a  `IntVarViewOpposite`, in which case we could test the value.

gostreap: We could just test the minimum and maximum actually.

3rdCore: I thought you were returning ` IntVarViewOpposite` in fact. I'm wondering what is the more logical way to think about it, at first sight, I'd say that `-` should generate a `IntVarViewOpposite`, we should consider a little practical case.

3rdCore: Recall that the goal of overloading the `-` operator is to facilitate the model declaration, this feature should generate an object that can be handled by the user. `IntDomainViewOpposite` is just a technical object relative to the operation of the `IntVarViewOpposite` variable.
 
```julia
    trailer = SeaPearl.Trailer()
    model = SeaPearl.CPModel(trailer)
    x = SeaPearl.IntVar(2, 3, "x", trailer)
    y = - x                 #y = SeaPearl.IntVarViewOpposite(x, "-x")
    SeaPearl.addVariable!(model, x)
    SeaPearl.addVariable!(model, y)
    ...
    SeaPearl.addConstraint!(model, SeaPearl.SumToZero([x, y], trailer))
```

PR #214, opened 2022-04-26
Optimization of the CP solver

marco-novaes98:
Using benchmarking and profiling tools to fix inefficiencies/bottlenecks to make the CP solver faster.

# Used tools:
- `ProfileView` to detect type instabilities (https://github.com/timholy/ProfileView.jl)
- `PProf` to visualize profiling data as an interactive flame graph to detect time-consuming functions (https://github.com/google/pprof/blob/master/doc/README.md)
- `BenchmarkTools`  to quantify the impact of the modifications on performance

# Benchmark results (using `@btime`): 
| Problem  | Instance | Benchmark before | Benchmark after | Relative reduction in execution time | 
| ------------- | ------------- | ------------- | ------------- | ------------- | 
| Graph coloring  | gc_50_3  | 1.035 s (5240443 allocations: 312.01 MiB) | 872.694 ms (5139583 allocations: 285.10 MiB)  | 15.7% |
| KEP  | kep_25_0.1  | 570.305 ms (2986995 allocations: 188.92 MiB) | 354.275 ms (2425328 allocations: 161.51 MiB)  | 37.9% |
|  | kep_30_0.1 | 22.897 s (117932471 allocations: 6.49 GiB) | 13.418 s (91215451 allocations: 5.78 GiB) | 41.4% |
| nqueens | size 9 | 682.374 ms (7244553 allocations: 548.39 MiB) |  679.452 ms (7237811 allocations: 548.04 MiB) | 0.4% |
| | size 10 | 2.634 s (27405222 allocations: 2.06 GiB) | 2.685 s (27349664 allocations: 2.05 GiB) | -1.9% | 
| | size 11 | 14.208 s (119028175 allocations: 9.10 GiB) | 12.394 s (109522814 allocations: 8.32 GiB) | 12.8% |
| TSPTW (seed=1) | size 30 | 6.034 s (73985081 allocations: 2.12 GiB) | 5.196 s (82641886 allocations: 2.40 GiB) | 13.9% |
| | size 40 | 19.903 s (305177539 allocations: 8.67 GiB) | 19.378 s (324116597 allocations: 9.17 GiB) | 2.7% |
| | size 50 | 88.197 s (1096370281 allocations: 31.93 GiB) | 72.691 s (1034126817 allocations: 30.07 GiB) | 17.6% |

# Analysis:
For the **KEP** instances we have the best results, reaching more than **40%**  reduction in execution time. This is probably due to the fact that the profiling analysis was carried out mainly using KEP and therefore the changes are somewhat optimised for this problem.

For **graph coloring** and **TSPTW** we have approximately a **15%** reduction in execution time. This shows that the modifications done for the KEP are useful for other problems even if to a lesser degree.

For **nqueens** the results are not conclusive. One explanation could be that for this problem there is no objective function and all the variables are branchable while some of the optimizations done are related to these elements.

# Flame graph example

Flame graph for the TSPTW (size 50) after optimization:
Captura de pantalla 2022-04-28 a las 10 26 55

From this graph we can get some interesting information about time spent inside functions:

- `solve_tsptw`: 96.1% 
  - `restoreState!`: 16%
  - `expandDfs!`: 79.2%
    - `fixPoint!`: 78.2%
     - `propagate!` (Element2D): 28.7%
     - `propagate!` (SumToZero): 14.2%
     - `propagate!` (BinaryMaximumBC): 8.7%
     - `propagate!` (ReifiedInSet): 8.7%
     - `propagate!` (SetDiffSingleton): 7.6%
     - `propagate!` (Equal): 4.7%
     - `propagate!` (isLessOrEqual): 3%
     - `propagate!` (InSet): 1.9%


src/CP/core/cpmodification.jl
```diff
@@ -19,7 +19,7 @@ function updateChildren!(prunedDomains::CPModification, x::Union{IntVar, BoolVar
     for y in x.children
         childrenPruned = childrenValue.([y], pruned)
         if haskey(prunedDomains, y.id)
-            prunedDomains[y.id] = vcat(prunedDomains[y.id], Array(childrenPruned))
+            prunedDomains[y.id] = append!(prunedDomains[y.id], Array(childrenPruned))
```
3rdCore, line 22: I didn't know that _vcat_ what not the best practice in this case. 

Just to know, what is the core difference between both methods ?

3rdCore (approved):
Nice work ! 

In the long run, it would be nice to completely remove the object `branchable` of the `CPModel`.

src/CP/core/cpmodification.jl
marco-novaes98: The performance difference is small but `vcat` creates a new array, `append!` modifies the array. So in this case the best practice is append! (see https://discourse.julialang.org/t/vcat-vs-push-which-one-is-more-efficient/48353). But you make me realize that the left side of the equal is not necessary, I will change that.

src/CP/core/utils/stopwatch.jl
```diff
@@ -7,6 +7,7 @@ Start a timer.
 function tic()
     t0 = time_ns()
     task_local_storage(:TIMERS, (t0, get(task_local_storage(), :TIMERS, ())))
+    nothing
```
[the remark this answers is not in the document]
marco-novaes98: This function was type unstable because Julia was not able to infer the returned type. As this function is not supposed to return something, I just added it explicitly to make life a little bit easier to the compiler even though I I guess the performance impact is almost null.

PR #218, opened 2022-05-02
Add RandomHeuristic for value

### LINKED ISSUE #213: Add a random value selection heuristic.
It is very common to compare a heuristic to a random heuristic. Currently, almost every experiment in SeaPearlZoo defines its own randomized heuristic by copying the code of another experiment.

So it would be interesting to add a random value heuristic to SeaPearl.

gostreap:
Random value heuristic to replace those written in each SeaPearlZoo experiment.

Resolve #213.


louis-gautier (approved):
Thank you for this useful new feature!

PR #220, opened 2022-05-04
RB generator

gostreap:
Adds a generator for RB problem as defined [here](https://www.sciencedirect.com/science/article/pii/S0004370207000653).

This should allow us to compare our performances with those obtained in [this article](https://arxiv.org/abs/1912.10762).


gostreap:
Because of a bad manipulation, this pull request has been duplicated, the review is in #217.

PR #223, opened 2022-05-04
Enable imitation learning

ziadelassal:
This PR adds an imitation learning functionality in a new `SupervisedLearnedHeuristic`, which is an adaptation of `LearnedHeuristic` (which name has changed to `SimpleLearnedHeuristic`).
1. At each `InitializingPhase`, ie at the beginning of each episode, and with a probability `eta`, the `SupervisedLearnedHeuristic` solves the instance using classic CP and adds the solution to its field `helpSolution`, which is set to `nothing` by default.
2. At each `DecisionPhase`:
i. if `helpSolution` is `nothing` the agent takes an action following its policy: `action = valueSelection.agent(env)`,
ii. if `helpSolution` contains a solution, the agent takes an action corresponding to the solution. At the end of the episode, the CP-obtained solution is reconstituted.

This approach is inspired from Gasse et al. ([https://arxiv.org/abs/1906.01629](https://arxiv.org/abs/1906.01629)) and allows the agent to learn from "good" examples even since the beginning of its training.


gostreap (changes requested):
Good job in general and well commented, I think it fits well in the solver. 

Simply, it is possible to avoid the duplication of functions common to `SimpleLearnedHeuristic` and `SupervisedLearnedHeuristic`. I will add this in a later commit.

It would be nice to add tests in order to check that the `helpSolution` is correctly set in addition to testing the reproducibility with `rng`.

gostreap (approved):

Project.toml
```diff
@@ -32,7 +32,7 @@ Flux = "0.11 - 0.12"
 JuMP = "0.21"
 LightGraphs = "1.3"
 MathOptInterface = "0.9"
-NNlib = "0.7"
+NNlib = "0.8.4"
```
louis-gautier, line 35: Useful to work with ReinforcementLearning.jl in dev mode

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,143 @@
...
+"""
+    (valueSelection::SupervisedLearnedHeuristic)(::DecisionPhase, model::CPModel, x::Union{Nothing, AbstractIntVar}, current_status::Union{Nothing, Symbol})
+
+Observe, store useful informations in the buffer with agent(POST_ACT_STAGE, ...) and take a decision with the call to agent(PRE_ACT_STAGE).
+The metrics that aren't updated in the StepPhase, which are more related to variables domains, are updated here. Once updated, they can
+be used in the other set_reward! function as the reward of the last action will only be collected in the RL.POST_ACT_STAGE.
+"""
+function (valueSelection::SupervisedLearnedHeuristic)(PHASE::Type{DecisionPhase}, model::CPModel, x::Union{Nothing,AbstractIntVar})
+    # domain change metrics, set before reward is updated
+    set_metrics!(PHASE, valueSelection.search_metrics, model, x)
```
louis-gautier, line 94: Are we sure we want the metrics to be totally independent from the chosen value heuristic type?

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    step = valueSelection.step
+    if step <= valueSelection.warmup_steps
+        return valueSelection.eta_init
+    elseif step >= (valueSelection.warmup_steps + valueSelection.decay_steps)
+        return valueSelection.eta_stable
+    else
+        steps_left = valueSelection.warmup_steps + valueSelection.decay_steps - step
+        return valueSelection.eta_stable + steps_left / valueSelection.decay_steps * (valueSelection.eta_init - valueSelection.eta_stable)
```
louis-gautier, line 141: Do we want a linear decay like this or would an exponential one be more adapted?

test/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,501 @@
...
+            eta_init=1.0,
+            eta_stable=0.0,
+            warmup_steps=2,
+            decay_steps=2
+        )
+        SeaPearl.update_with_cpmodel!(lh, model)
+        false_x = first(values(model.variables))
+        env = SeaPearl.get_observation!(lh, model, false_x)
+
+
```
louis-gautier, line 98: Perfect to do step-by-step tests like this. Thank you very much for the time spent on this. It will ensure that the behavior of `SupervisedLearnedHeuristic` is as expected.

louis-gautier (approved):

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,143 @@
...
+        strategy = DFSearch()
+        search!(model_duplicate, strategy, valueSelection.helpVariableHeuristic, valueSelection.helpValueHeuristic)
+        #the solution is added to the valueSelection.helpSolution field.
+        if !isnothing(model_duplicate.statistics.solutions)
+            solutions = model_duplicate.statistics.solutions[model_duplicate.statistics.solutions.!=nothing]
+            if length(solutions) >= 1
+                valueSelection.helpSolution = solutions[1]
+            end
+        end
+        reset_model!(model_duplicate)
```
3rdCore, line 72: Why do we reset the duplicated model ?

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,143 @@
...
+    if rand(valueSelection.rng) < eta
+        #the instance is solved using classic CP on a duplicated model
+        model_duplicate = deepcopy(model) 
+        strategy = DFSearch()
+        search!(model_duplicate, strategy, valueSelection.helpVariableHeuristic, valueSelection.helpValueHeuristic)
+        #the solution is added to the valueSelection.helpSolution field.
+        if !isnothing(model_duplicate.statistics.solutions)
+            solutions = model_duplicate.statistics.solutions[model_duplicate.statistics.solutions.!=nothing]
+            if length(solutions) >= 1
+                valueSelection.helpSolution = solutions[1]
```
3rdCore, line 69: Why do we select the first solution found, instead of the last solution found which is supposed to be of better quality!

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,143 @@
...
+            end
+        end
+        reset_model!(model_duplicate)
+    else
+        valueSelection.helpSolution = nothing
+    end
+
+    # Reset the agent, useful for things like recurrent networks
+    Flux.reset!(valueSelection.agent)
```
3rdCore, line 78: I'm just wondering why we previously added this line?

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,143 @@
...
+
+    return action_to_value(valueSelection, action, state(env), model)
+end
+
+"""
+    get_eta(valueSelection::SupervisedLearnedHeuristic)
+
+Get the current value of `eta` (η), which is the probability for the solver to calculate and provide a classic CP-generated solution to the agent.
+"""
+function get_eta(valueSelection::SupervisedLearnedHeuristic)
```
3rdCore, line 128: To what extend do we need equation to have its own specific decay rate ? 

Could we have used the equation value of the  `agent.policy.explorer` ?

test/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,501 @@
...
+        @test !isnothing(lh.helpSolution)
+        @test lh.helpSolution == Dict{String,Union{Bool,Int64,Set{Int64}}}("x1" => 2, "x4" => 4, "x2" => 1, "x3" => 3)
+
+        # We check that the values returned by the heuristic correspond to those of the helpSolution.
+        # Variable 1
+        lh(SeaPearl.StepPhase, model, :Feasible)
+        x = variable_heuristic(model)
+        v = lh(SeaPearl.DecisionPhase, model, x)
+        @test v == lh.helpSolution[x.id]
+        lh(SeaPearl.EndingPhase, model, :Feasible)
```
3rdCore, line 307: Is this line usefull?

test/CP/valueselection/learning/supervisedlearnedheuristic.jl
```diff
@@ -0,0 +1,501 @@
...
+        lh(SeaPearl.InitializingPhase, model)
+        @test isnothing(lh.helpSolution)
+        SeaPearl.reset_model!(model)
+        
+        lh(SeaPearl.InitializingPhase, model)
+        @test !isnothing(lh.helpSolution)
+        SeaPearl.reset_model!(model)
+        
+        lh(SeaPearl.InitializingPhase, model)
+        @test isnothing(lh.helpSolution)
```
3rdCore, line 498: Nice check using the `rng`.

3rdCore (approved):

test/CP/valueselection/learning/supervisedlearnedheuristic.jl
gostreap: We have tried to keep the heuristic calls as close as possible, in order to detect possible problems in the future. However, I don't think that it will influence the result of the test in the current state of the removed. 

I am however in favor of keeping this line in the test.

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
gostreap: I honestly have no idea, it seemed to me that this was not a problem but I don't know if it is useful.

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
gostreap: This could indeed be a much better approach!

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
gostreap: This is probably unnecessary and should indeed be removed.

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
gostreap: I'm not sure I understand, what would you consider here?

src/CP/valueselection/learning/supervisedlearnedheuristic.jl
gostreap: It would indeed be nice to be able to choose, but it doesn't seem to me to be absolutely essential at the moment.

gostreap:
From the last comments, it seems that we have merged the branch a bit early. I will work on a new small pull request to correct the few points mentioned.

PR #226, opened 2022-05-06
Creating a general reward encouraging a smart exploration of the tree

louis-gautier:
This PR adds a reward called `ExperimentalReward` (will have to be renamed to `GeneralReward` once merged with master).

The goal of this reward is to be as general as possible to encourage a smart exploration of the tree of solutions of any problem solved using `DefaultStateRepresentation`. Its functioning is briefly exposed below:

* By design, we make sure that any found feasible solution always gets a higher reward than any infeasible solution.
* We choose to assign rewards to each decision and not only in `EndingPhase` as it was previously done with `CPReward`. It should allow faster learning in large search trees and foster a smart exploration of this tree.

The rewards are given as described here:

* **In all cases:**

We encourage the agent to bound as many variables as possible in one assignation. The agent thus receives a reward of equation (with equation to make this function convex) at each decision.

At the `EndingPhase`, the agent receives a negative penalty of equation if it reaches an infeasible solution.

* **If the problem has an objective function**, we add the following elements:

We encourage the agent to keep low values in the domain of the objective function throughout the resolution. The agent thus receives a reward of equation.

At the `EndingPhase`, the agent receives a negative penalty of -1 if it reaches an infeasible solution.


3rdCore:
That would be nice to have some comparison with other reward know to be working on specific problems ( tsptw, graphColoring ..) before merging what do you think of this ?

louis-gautier:
> That would be nice to have some comparison with other reward know to be working on specific problems ( tsptw, graphColoring ..) before merging what do you think of this ?

Sure. Here is a comparison of `GeneralReward` ("learned") and `CPReward` ("learned2") on graphcoloring(20) with the default hyperparameters for both rewards.

c35d3fc0-71b6-40ad-939d-d261bfa7b780

On TSPTW(10) however, no learning can be observed (the rewards are consistent but the agent does not learn), probably due to pitfalls in the treatment of the representation as the graph is very large on TSPTW.

@malikattalah has worked on comparing the 2 rewards on nqueens.

src/CP/core/search/dfs.jl
```diff
@@ -52,6 +77,7 @@ function expandDfs!(toCall::Stack{Function}, model::CPModel, variableHeuristic::
     x = variableHeuristic(model)
     # Value selection
     v = valueSelection(DecisionPhase, model, x)
+    #println("value: "*string(v))
```
gostreap, line 80: We can delete the commented prints in this file.

src/CP/valueselection/learning/rewards/generalreward.jl
```diff
@@ -0,0 +1,122 @@
+"""
+    struct GeneralReward <: AbstractReward end
+
+This is a general reward encouraging a smart exploration of the tree in the case DFSearch is used.
```
gostreap, line 4: I think it would be nice to add some comments on the general design of the reward, in the same way as it was done in the description of the pull request.

Like for example that 
> By design, we make sure that any found feasible solution always gets a higher reward than any infeasible solution.

src/CP/valueselection/learning/rewards/generalreward.jl
```diff
@@ -0,0 +1,122 @@
...
+}
+    if !lh.firstActionTaken
+        #println("First action being taken")
+        lh.reward.value = 0
+    else
+        lh.reward.value = lh.reward.gamma*(model.statistics.lastPruning/(lh.reward.initialNumberOfVariableValueLinks - length(branchable_variables(model))))^lh.reward.beta
+        #println(model.statistics.lastPruning)
+        #println(lh.reward.initialNumberOfVariableValueLinks - length(branchable_variables(model)))
+        #println(string(lh.reward.value))
+        if lh.reward.value>0.001
```
gostreap, line 53: This `if` is now unnecessary.

src/CP/valueselection/learning/rewards/generalreward.jl
```diff
@@ -0,0 +1,122 @@
...
+    set_reward!(::EndingPhase, lh::LearnedHeuristic{GeneralReward, A}, model::CPModel, symbol::Union{Nothing, Symbol})
+
+Increment the current reward at the EndingPhase. Called when the search is finished by an optimality proof or by a limit in term of nodes or 
+in terms of number of solution. This is useful to add some general results to the reward like the number of ndoes visited during the episode for instance. 
+"""
+function set_reward!(::Type{EndingPhase}, lh::LearnedHeuristic{SR, GeneralReward, A}, model::CPModel, symbol::Union{Nothing, Symbol}) where {
+    SR <: AbstractStateRepresentation, 
+    A <: ActionOutput
+}
+    nqueens_conflict_counter = false
```
gostreap, line 80: This field is now useless as far as I can tell because never updated and seems to me to be inherited from an experimental reward proper to nqueens.

src/experiment/metrics/basicmetrics.jl
```diff
@@ -94,8 +113,19 @@ function (metrics::BasicMetrics{TakeObjective, <:LearnedHeuristic})(model::CPMod
...
+    total_reward = last_episode_total_reward(metrics.heuristic.agent.trajectory)
+    #=
+    # Uncomment to print these metrics during training
+    print("Reward: ")
+    println(total_reward)
+    print("Loss: ")
+    println(metrics.heuristic.agent.policy.learner.loss)
+    println("")
+    =#
+    push!(metrics.totalReward,total_reward)
```
gostreap, line 128: In the same way these changes do not seem to me to be merge.

src/experiment/metrics/basicmetrics.jl
```diff
@@ -94,8 +113,19 @@ function (metrics::BasicMetrics{TakeObjective, <:LearnedHeuristic})(model::CPMod
     index = findfirst(!isnothing, model.statistics.solutions) #return the list of index of real solution in model.statistics.solutions
     push!(metrics.meanNodeVisitedUntilfirstSolFound, !isnothing(index) ? model.statistics.nodevisitedpersolution[index] : nothing)
     push!(metrics.timeneeded,dt)
-    push!(metrics.scores,copy(model.statistics.objectives))
-    push!(metrics.totalReward,last_episode_total_reward(metrics.heuristic.agent.trajectory))
+    if !isempty(model.statistics.objectives)
```
gostreap, line 116: Did it cause an error to push an empty array? If not, I don't think this change is desirable as it could cause problems in the `SeaPearlExtras` plotting functions.

gostreap (changes requested):
The reward looks very well designed to me and will probably improve performance compared to previous general rewards. I look forward to this being merged into the main branch.

However before that, I think it is necessary to clean up the code a bit, especially by removing all prints that were only related to the development of the reward.

It would also be nice to add some tests to keep the project on the right track regarding unit tests.

src/CP/core/search/dfs.jl
```diff
@@ -35,7 +35,32 @@ function expandDfs!(toCall::Stack{Function}, model::CPModel, variableHeuristic::
...
+                    # Pruning from the middle of the domain of the objective variable
+                    model.statistics.objectiveDownPruning += (model.objective.domain.max.value - val)/(model.objective.domain.max.value-model.objective.domain.min.value)
+                    model.statistics.objectiveUpPruning += (val - model.objective.domain.min.value)/(model.objective.domain.max.value-model.objective.domain.min.value)
+                end
+            end
+        else
+            model.statistics.objectiveDownPruning = 0
+            model.statistics.objectiveUpPruning = 0
+        end
+    end
```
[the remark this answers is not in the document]
louis-gautier: Very good idea. I created a function `updateStatistics!` in `model.jl` for this purpose.

src/CP/valueselection/learning/rewards/generalreward.jl
louis-gautier: Very good idea to improve clarity. Done.

src/CP/valueselection/learning/rewards/generalreward.jl
louis-gautier: Was used for debugging only. All the instructions and comments used for debugging are now removed.

src/experiment/metrics/basicmetrics.jl
```diff
@@ -80,7 +80,26 @@ function (metrics::BasicMetrics{DontTakeObjective, <:LearnedHeuristic})(model::C
...
+        println("Network weights")
+        println(norm(metrics.heuristic.agent.policy.learner.approximator.model.graphChain[1].weight1))
+        println(norm(metrics.heuristic.agent.policy.learner.approximator.model.graphChain[2].weight1))
+        println(norm(metrics.heuristic.agent.policy.learner.approximator.model.nodeChain[1].weight))
+        println(norm(metrics.heuristic.agent.policy.learner.approximator.model.nodeChain[2].weight))
+        println(norm(metrics.heuristic.agent.policy.learner.approximator.model.outputChain[1].weight))
+        println(norm(metrics.heuristic.agent.policy.learner.approximator.model.outputChain[2].weight))
+    end
+    =#
+    push!(metrics.totalReward,total_reward)
```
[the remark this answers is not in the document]
louis-gautier: I simplified those prints but I still kept it as they can be very useful during development to do quick debugging.

src/experiment/metrics/basicmetrics.jl
louis-gautier: I did this because of an error (despite being aware that it might cause conflicts with `SeaPearlExtras` but now I don't get this error anymore when running graphcoloring so I removed it but we should be ready to reverse this change if we see an error related to this line in the future.

src/RL/nn_structures/cpnn.jl
```diff
@@ -53,3 +53,44 @@ function (nn::CPNN)(states::BatchedDefaultTrajectoryState)
         return output
     end
 end
+
+# Overloads the Base.string() function for storing parameters of the neural networks associated to experiments.
+function Base.string(nn::CPNN)
```
3rdCore, line 58: Do we really need this as long as CPNN will be partially refacto ?

3rdCore (approved):

3rdCore:
I totally agree with @gostreap regarding the testsets !

louis-gautier:
Unit tests have been added for this reward. We should soon be able to merge, which will help us conducting tests on problems with objective variables for the new heterogeneous graph pipeline.

src/RL/nn_structures/cpnn.jl
louis-gautier: It has nothing to do with the reward but it is really useful to keep track of the neural networks we use when we do experiments (the returned string is saved in `params.json` files). For performing experiments on the new heterogeneous pipeline we have duplicated CPNN so I think we should keep it like this for the moment.

gostreap (approved):

Issue #225, opened 2022-05-06
Refacto of the Graph Convolution


3rdCore:
## The Issue:  Comparing apple with oranges...

After a long discussion with all the interested parties, we came to the conclusion that the graph convolution pipeline applied on the default tripartite representation had some serious homogeneity problems. Indeed, the same aggregation operation was applied to nodes of different nature. 

<img src="https://user-images.githubusercontent.com/59280588/167154290-5c30fc26-51dd-40be-adf0-9448881ff4a6.png" width="200">

Conceptually, we were trying to compare apples with oranges... 

In SeaPearl, we encode conceptual CP problems in a tripartite graph representation that can be processed by our RL agent. In this representation, a node can represents either a **variable**, a **value**, or a **constraint**.
 
So far, for any node, no matter its nature, the same Graph convolution was applied : 

![equation](https://latex.codecogs.com/svg.image?\mathbf{h}_{u}^{t&space;&plus;1}=\phi\left(\mathbf{h}_{u}^t\:&space;&space;&space;\bigoplus_{v&space;\in&space;\mathcal{N}_{u}}&space;\mathbf{h}_{v}^t\\right)&space;)

Where  ![equation](https://latex.codecogs.com/svg.image?\phi&space;) is a **One-Layer Perceptron** : a learnable affine transformation + activation functions such that for any input vector x : ![equation](https://latex.codecogs.com/svg.image?\phi(x)&space;=&space;\sigma(Wx&space;&plus;&space;B&space;)). **Average Pooling, Mean Pooling, Max pooling** can be used for the aggregation function.

This operation can be computed efficiently across the entire set of nodes using matrix formulation which is particularly adapted for high-performance GPU computing, in case of a sum Pooling : 

![equation](https://latex.codecogs.com/svg.image?\mathbf{H}^{t&plus;1}&space;=&space;\sigma&space;(W(\mathbf{H}^{t}&space;:&space;A\mathbf{H}^{t})^T&space;&plus;&space;B))

Where  ![equation](https://latex.codecogs.com/svg.image?\mathbf{H}^{t}) are nodes features at time step _t_ (**col-wise**), ![equation](https://latex.codecogs.com/svg.image?W,&space;B) are learnable parameters  : ![equation](https://latex.codecogs.com/svg.image?\sigma) an activation function, and ![equation](https://latex.codecogs.com/svg.image?A) the adjacency matrix of the graph. 
 
[...]

PR #229, opened 2022-05-10
Heterogeneous graph and new associated RL pipeline.

gostreap:
The purpose of this pull request is to allow the implementation of the new graph convolution discussed in issue #225.

The major novelties are :
- Introduction of a heterogeneous graph with 3 distinct node types for constraints, variables and values.
- The modification of the tripartite graph to better manage `VarViews`:
   `VarView` relations were formerly represented as an edge between 2 distinct variables (the parent and its child). With this PR, `VarView` relations are now represented as constraints of type `ViewConstraint` linked to these 2 variables. When `constraint_type` is set to `true` in `chosen_features`, and a problem containts `VarView` relations, it now encodes the type of the constraint with 4 new dimensions reflecting the exact nature of the `VarView` relation. Please refer to `featurize` function in `heterogeneousstaterepresentation.jl` for more information.
- The implementation of the new convolutions as described in #225.
  - The initialization layer of the convolution :  `HeterogeneousGraphConvInit`
  - The heterogeneous version of `GraphConv`: `HeterogeneousGraphConv`.
- The compatibility of all downstream neural networks (CPNNs) with this new heterogeneous pipeline: `HeterogeneousCPNN`, `HeterogeneousFullFeaturedCPNN` and `HeterogeneousVariableOutputCPNN`.
- The resolution of major bugs in the pipeline:
  - Fix of model emptying after the resolution of each instance
  - Fix of value ordering in the graph construction
- The implementation of a new `mean` aggregation in the convolutional layers.

Minor updates also concern:
- The removal of symmetries in the generation of graphcoloring problems
- The correction of errors in the generation of KEP and RB problems


qcappart:
Great work ! I am very impressed by the rapidity to implement this. I am very eager to see the results ! :-)

gostreap:
It is not yet fully functional, but we are opening a draft to allow everyone to see the progress and to make comment ! :)

louis-gautier:
The heterogeneous graph pipeline is now running without bugs, but we still have to write some unit tests to ensure that its behavior is as expected.

src/RL/representation/default/cp_layer/types.jl
```diff
@@ -121,6 +120,14 @@ struct CPLayerGraph <: LightGraphs.AbstractGraph{Int}
             nodeToId[ConstraintVertex(cpmodel.constraints[i])] = i
         end
 
+        # Filling view constraints
+        for i in 1:length(viewConstraints)
+            idToNode[numberOfConstraints + i] = ConstraintVertex(viewConstraints[i])
+            nodeToId[ConstraintVertex(viewConstraints[i])] = numberOfConstraints + i
+        end
```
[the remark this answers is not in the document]
3rdCore: I think I got a better understanding of `viewConstraints`. These are a type of constraints to specify that a variable is a view (of any kind) of another variable ?

3rdCore:
Implemented `HeterogeneousVariableOutputCPNN`. I'll add `HeterogeneousFullFeaturedCPNN` on monday. I added the field `possibleValuesIdx` to the `DefaultTrajectoryState` to match `HeterogenousTrajectoryState` fields. 

Many more testsets will be added next week to ensure that the field `possibleValuesIdx` is working as expected.

src/RL/representation/default/heterogeneousstaterepresentation.jl
```diff
@@ -90,6 +91,9 @@ Update the StateRepesentation according to its Type and Featurization.
 """
 function update_representation!(sr::HeterogeneousStateRepresentation, model::CPModel, x::AbstractIntVar)
     update_features!(sr, model)
+    ncon = sr.cplayergraph.numberOfConstraints
+    nvar = sr.cplayergraph.numberOfVariables
+    sr.possibleValuesIdx = map(v -> indexFromCpVertex(sr.cplayergraph, ValueVertex(v)), collect(x.domain)) .- ncon .- nvar
```
3rdCore, line 96: I definitely need your approval on this line.

louis-gautier: This line is correct! However, some errors are raised when we compute `valueFeatures` from `possibleValuesIdx` in the case of a `BatchedHeterogeneousTrajectoryState`. I am currently working on it.

src/RL/nn_structures/heterogeneousfullfeaturedcpnn.jl
```diff
@@ -21,36 +21,27 @@ Flux.@functor HeterogeneousFullFeaturedCPNN
...
-    variableFeature = fg.varnf[:, variableIdx]
-    valueFeatures = fg.valnf[:, possibleValuesIdx]
+    variableFeatures = featuredGraph.varnf # FxNxB
+    valueFeatures = featuredGraph.valnf
     globalFeatures = featuredGraph.gf # GxB
 
     # Extract the features corresponding to the varibales
     variableIndices = nothing
     Zygote.ignore() do
         variableIndices = Flux.unsqueeze(CartesianIndex.(variableIdx, 1:batchSize), 1)
```
3rdCore, line 38: I know that this hasn't been reconsidered for a while but I don't understand why we have to transform the Cartesian Index into a 1-row matrix of type : `1×4 Matrix{CartesianIndex{2}}`.

3rdCore: Well, it is just a matter of dimensions :

Here is a little dummy example done on a handcrafted `batchedDefaultTrajectoryState `: 
```julia
julia> idx = CartesianIndex.(zip([3,4], 1:2))
2-element Vector{CartesianIndex{2}}:
 CartesianIndex(3, 1)
 CartesianIndex(4, 2)

julia> idx_unsqueeze = Flux.unsqueeze(CartesianIndex.(zip([3,4], 1:2)),1)
1×2 Matrix{CartesianIndex{2}}:
 CartesianIndex(3, 1)  CartesianIndex(4, 2)

julia> batched_state_2.fg.nf[:,idx]
3×2 Matrix{Float32}:
 0.0  0.0
 1.0  1.0
 0.0  0.0

julia> batched_state_2.fg.nf[:,idx_unsqueeze]
3×1×2 Array{Float32, 3}:
[:, :, 1] =
 0.0
 1.0
 0.0

[:, :, 2] =
 0.0
 1.0
 0.0
```

3rdCore (changes requested):

src/RL/representation/default/defaultstaterepresentation.jl
```diff
@@ -113,7 +113,7 @@ function update_representation!(sr::DefaultStateRepresentation, model::CPModel,
     update_features!(sr, model)
     ncon = sr.cplayergraph.numberOfConstraints
     nvar = sr.cplayergraph.numberOfVariables
-    sr.possibleValuesIdx = map(v -> indexFromCpVertex(sr.cplayergraph, ValueVertex(v)), collect(x.domain)) .- ncon .- nvar
+    sr.possibleValuesIdx = map(v -> indexFromCpVertex(sr.cplayergraph, ValueVertex(v)), collect(x.domain))
```
3rdCore, line 116: I had to remove my change because it wasn't working anymore with `variableoutputcpnn` (index issue raised at [line 32](https://github.com/corail-research/SeaPearl.jl/blob/heterogeneous/src/RL/nn_structures/variableoutputcpnn.jl)).

I modified the function `from_order_to_id` consequently.

src/RL/nn_structures/heterogeneousfullfeaturedcpnn.jl
```diff
@@ -21,36 +21,27 @@ Flux.@functor HeterogeneousFullFeaturedCPNN
...
     end
-    variableFeatures = nodeFeatures[:, variableIndices] # Fx1xB
-    variableFeatures = reshape(nn.nodeChain(RL.flatten_batch(variableFeatures)), :, 1, batchSize) # F'x1xB
+    branchingVariableFeatures = variableFeatures[:, variableIndices] # Fx1xB
+    relevantVariableFeatures = reshape(nn.nodeChain(RL.flatten_batch(branchingVariableFeatures)), :, 1, batchSize) # F'x1xB
```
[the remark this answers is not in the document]
louis-gautier: We could potentially try it but I think we should limit the number of weights in our neural networks:  `HeterogeneousGraphConv` already deals differently with variables and values, so it might be unnecessary to further maintain differences in treatment.

src/RL/nn_structures/heterogeneousfullfeaturedcpnn.jl
louis-gautier: Thank you for this example!

src/RL/representation/default/defaultstaterepresentation.jl
louis-gautier: Ok, no problem if it is adapted in `from_order_to_id`

src/RL/nn_structures/heterogeneousfullfeaturedcpnn.jl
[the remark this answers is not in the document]
3rdCore: Well, the pipeline must be heterogeneous at every step, otherwise, all the work done to increase representation on the convolution will be bottlenecked by this inconsistent layer.  Let's discuss that with the others !

@gostreap @malikattalah @ziadelassal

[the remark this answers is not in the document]
3rdCore: I have another comment, what if the dimensions of the variable embedding and the value embedding do not match? This requires that the GNN satisfy this condition.

louis-gautier:
> Nice bug spot ! That will indeed simplificate the tripartite graph representation !

I still have to look at the testsets for graph coloring problem generation as tests are now failing.

src/RL/representation/default/heterogeneoustrajectorystate.jl
```diff
@@ -0,0 +1,109 @@
...
+"""
+struct BatchedHeterogeneousTrajectoryState{T} <: NonTabularTrajectoryState
+    fg::BatchedHeterogeneousFeaturedGraph{T}
+    variableIdx::AbstractVector{Int}
+    possibleValuesIdx::Union{Nothing, AbstractVector{Vector{Int64}}} #Todo change this to abstract matrix
+
+function BatchedHeterogeneousTrajectoryState(fg, var, val)
+    @assert all(var .<= size(fg.varnf,2)) "The variable index is out of bound"
+    @assert all(reduce(vcat,val) .<= size(fg.valnf,2)) "One of the values index is out of bound"
+    @assert all(reduce(vcat,[[tuple[2] for tuple in countmap([c for c in vals])] for vals in val]) .== 1) "values array contains one element at least twice"
```
3rdCore, line 40: Well, This is absolutely inefficient and must be removed for production, but I'd prefer having these kinds of verifications to prevent any anti-pattern behavior while trying to make our new pipeline work.

src/RL/utils/geometricflux/heterogeneousgraphconvinit.jl
```diff
@@ -0,0 +1,43 @@
...
+    weightsvar::A
+    weightscon::A
+    weightsval::A
+    biasvar::B
+    biascon::B
+    biasval::B
+    σ
+end
+
+function HeterogeneousGraphConvInit(in::Array{Int}, out::Int, σ=identity; init=Flux.glorot_uniform, bias::Bool=true, T::DataType=Float32)
```
3rdCore, line 11: The base function is `σ=identity` following what has been done on [this paper.](https://www.sciencedirect.com/science/article/pii/S0952197621004255?via%3Dihub)

src/RL/representation/default/heterogeneousstaterepresentation.jl
```diff
@@ -226,6 +234,16 @@ function initChosenFeatures!(sr::HeterogeneousStateRepresentation{DefaultFeaturi
...
 
+        if haskey(chosen_features, "variable_is_objective") && chosen_features["variable_is_objective"]
+            sr.chosenFeatures["variable_is_objective"] = (true, variable_counter)
+            variable_counter += 1
+        end
+
+        if haskey(chosen_features, "nb_not_bounded_variable") && chosen_features["nb_not_bounded_variable"]
+            sr.chosenFeatures["nb_not_bounded_variable"] = (true, variable_counter)
+            variable_counter += 1
+        end
```
3rdCore, line 245: @louis-gautier, this 4 lines are very close from line 246 to 249, I'm wondering if this is a copy/paste mistake ?

3rdCore: It was.

src/CP/constraints/constraints.jl
```diff
@@ -65,4 +65,12 @@ triggerDomainChange!(toPropagate::Set{<:Constraint}, x::Union{AbstractIntVar, Ab
...
 
-variablesArray(constraint::Constraint) = throw(ErrorException("missing function variablesArray(::$(typeof(constraint)))."))
\ No newline at end of file
+variablesArray(constraint::Constraint) = throw(ErrorException("missing function variablesArray(::$(typeof(constraint)))."))
+
+
+struct ViewConstraint <: Constraint
+    parent  ::Union{SeaPearl.AbstractIntVar, SeaPearl.AbstractBoolVar}
+    child   ::Union{SeaPearl.IntVarView, SeaPearl.BoolVarView}
+end
```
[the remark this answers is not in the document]
gostreap: The role of this structure is to allow to represent in the graph a link between a variable and associated view.

For example, if we have a `VarView` `y = x + 3`, in the graph we will have two variable nodes which will be linked together by a `ViewConstraint`.

Previously there was an edge between the two variables, which caused several problems.

src/RL/utils/geometricflux/graphconv.jl
```diff
@@ -1,23 +1,28 @@
...
     b = bias ? T.(init(ch[2])) : zeros(T, ch[2])
-    return GraphConv(W1, W2, b, σ)
+    return GraphConv(W1, W2, b, σ, pool)
 end
 
 Flux.@functor GraphConv
 
 function (g::GraphConv)(fgs::BatchedFeaturedGraph{Float32}) 
     A, X = fgs.graph, fgs.nf
+    if isa(g.pool,meanPooling)
```
gostreap, line 23: The convolutions should be compatible with sum pooling. Eventually, passing through parametric types with different functions would be a solution.

That is, write two functions that are more or less typed like this:
- `function (g::GraphConv{A, B, meanPooling})(fgs::BatchedFeaturedGraph{Float32})`
- `function (g::GraphConv{A, B, sumPooling})(fgs::BatchedFeaturedGraph{Float32})`

This would avoid ifs in the function, which I think is not easy to handle with Zygote.

gostreap (commented):
In the current state the convolutions with mean pooling do not work. This must be fixed absolutely before the merge.

gostreap:
Pooling should work now. For `meanPooling`, the `Zygote.ignore()` covers a lot of things, we'll have to watch that it doesn't break the backpropagation.

@3rdCore, can you take a look at my modifications?

src/RL/representation/default/defaultstaterepresentation.jl
```diff
@@ -188,6 +190,9 @@ function featurize(sr::DefaultStateRepresentation{DefaultFeaturization,TS}; chos
             if sr.chosenFeatures["variable_is_objective"][1]
                 features[sr.chosenFeatures["variable_is_objective"][2], i] = sr.cplayergraph.cpmodel.objective == cp_vertex.variable
             end
+            if sr.chosenFeatures["variable_assigned_value"][1]
+                features[sr.chosenFeatures["variable_assigned_value"][2], i] = isbound(cp_vertex.variable) ? cp_vertex.variable.value : 0
```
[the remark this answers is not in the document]
louis-gautier: We don't have any control on the values that are assigned to variables. That's why `variable_assigned_value` always has to be used with `variable_is_bound`, such as the agent can differentiate a situation where the variable is assigned to 0 and a situation where the variable has not been assigned yet.

src/RL/utils/geometricflux/heterogeneousgraphconv.jl
```diff
@@ -68,4 +54,50 @@ function (g::HeterogeneousGraphConv)(fg::HeterogeneousFeaturedGraph, original_fg
...
+
+        return BatchedHeterogeneousFeaturedGraph{Float32}(
+            contovar,
+            valtovar,
+            g.σ.(g.weightsvar ⊠ vcat(X1, H1, H2 ⊠ contovarN, H3 ⊠ valtovarN) .+ g.biasvar),
+            g.σ.(g.weightscon ⊠ vcat(X2, H2, H1 ⊠ vartoconN) .+ g.biascon),
+            g.σ.(g.weightsval ⊠ vcat(X3, H3, H1 ⊠ vartovalN) .+ g.biasval),
+            fgs.gf
+        )
+    end
```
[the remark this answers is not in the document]
louis-gautier: We should not forget to correct this before merging with master. As the function is not recursive, it is ok to return under `Zygote.ignore()` statement but the operations that are computed inside the return here should be copied outside thr `Zygote.ignore()` domain. We have solved the same issue on `heterogeneousgraphtransformer.jl`.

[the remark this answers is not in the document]
louis-gautier: What causes the Zygote errors is the construction of the `BatchedHeterogeneousFeaturedGraph`. It would even be safer to do separate construction of the object under Zygote.ignore() and return outside it.

src/RL/representation/default/cp_layer/types.jl
```diff
@@ -81,7 +81,7 @@ struct CPLayerGraph <: LightGraphs.AbstractGraph{Int}
     """
     function CPLayerGraph(cpmodel::CPModel)
         variables = Set{AbstractVar}(values(cpmodel.variables))
-        valuesOfVariables = branchable_values(cpmodel)
+        valuesOfVariables = sort(branchable_values(cpmodel))
```
3rdCore, line 84: can you explain briefly what the issue was with the previous pipeline ?

louis-gautier: In short, the ordering of the actions taken by the agent was not consistent with the ordering of value node features in the graph. As a result, the values embeddings in `FullFeaturedCPNN` didn't match the actions taken by the agent afterwards. But this fix doesn't impact other CPNNs.

src/CP/core/model.jl
```diff
@@ -257,6 +257,8 @@ Empty the CPModel.
 """
 function Base.empty!(model::CPModel)
     empty!(model.variables)
+    empty!(model.branchable_variables)
+    empty!(model.branchable)
```
louis-gautier, line 261: These lines fix a major bug caused by the fact that the model was not properly emptied after the resolution of each instance in the training process.

src/CP/core/model.jl
```diff
@@ -364,17 +366,26 @@ end
...
+                end
+            end
+        elseif isa(x.domain,IntSetDomain)
+            cardinality += length(x.domain)
+        else
+            cardinality += length(x.domain)
+            if !isempty(x.children)
+                for child in x.children
+                    cardinality += length(child.domain)
+                end
```
louis-gautier, line 388: We now take all variables, including `VarViews` into consideration in `global_domain_cardinality`

src/RL/representation/default/defaultstaterepresentation.jl
```diff
@@ -149,6 +158,19 @@ function featurize(sr::DefaultStateRepresentation{DefaultFeaturization,TS}; chos
...
+                    elseif isa(cp_vertex.constraint.child, IntVarViewOffset)
+                        features[sr.constraintTypeToId[typeof(cp_vertex.constraint)] + 2, i] = cp_vertex.constraint.child.c
+                    elseif isa(cp_vertex.constraint.child, IntVarViewOpposite)
+                        features[sr.constraintTypeToId[typeof(cp_vertex.constraint)] + 1, i] = -1
+                    elseif isa(cp_vertex.constraint.child, BoolVarViewNot)
+                        features[sr.constraintTypeToId[typeof(cp_vertex.constraint)] + 3, i] = 1
+                    else
+                        error("WARNING: Unknwon VarViewType: please implement DefaultFeaturization for this type!")
+                    end
+                end
```
louis-gautier, line 173: If the problem we are trying to solve has `ViewConstraints`, in `constraint_type`, we encode them thanks to 4 new components, respectively: one-hot encoding of the fact the considered constraint is a `ViewConstraint`, multiplication coefficient if the `ViewConstraint` is of type `IntVarViewMul` or `IntVarViewOpposite`, offset coefficient if the if the `ViewConstraint` is of type `IntVarViewOffset`, and one-hot encoding of the fact the `ViewConstraint` is of type `BoolVarViewNot`.

src/datagen/coloring.jl
```diff
@@ -94,8 +95,8 @@ function fill_with_generator!(cpmodel::CPModel, gen::HomogenousGraphColoringGene
     
     # edge constraints
     for i in 1:n
-        for j in 1:n
-            if i != j && rand(rng) <= p
+        for j in (i+1):n
```
louis-gautier, line 98: This helps to remove symmetries when generating graphcoloring instances.

louis-gautier (changes requested):
We just have to check that `Zygote.ignore()` statements in all convolutional layers are well positioned and we can merge this branch on `master`.

src/CP/core/model.jl
3rdCore: Nice spot !

src/RL/utils/geometricflux/graphconv.jl
```diff
@@ -43,28 +43,34 @@ end
...
     Zygote.ignore() do
-        A = A ./ reshape(mapslices(x -> sum(eachrow(x)), A, dims=[1, 2]), 1, :, size(A, 3))
-
-        return BatchedFeaturedGraph{Float32}(
-            fgs.graph;
-            nf=g.σ.(g.weight1 ⊠ X .+ g.weight2 ⊠ X ⊠ A .+ g.bias),
-            ef=fgs.ef,
-            gf=fgs.gf
-        )
+        sum = replace(reshape(mapslices(x -> sum(eachrow(x)), A, dims=[1, 2]), 1, :, size(A, 3)), 0=>1)
```
3rdCore, line 48: What is the reason for adding a `replace( M, 0=>1)` ?

3rdCore: edit : This is to avoid any division by 0, in case a node has no neighbors, which can happen for value that belongs to no domain after pruning.

gostreap: In the case where a node has no neighbor, the sum is zero, which will then cause problems in the division.

Putting a 1 in the division should not be a problem, because it will divide a vector composed only of 0. Any constant other than 1 would have done the job as well.

test/datagen/coloring.jl
```diff
@@ -13,7 +13,7 @@
         SeaPearl.fill_with_generator!(model, generator; rng = rng)
 
         @test length(keys(model.variables)) == nb_nodes + 1
-        @test length(model.constraints) == 22
+        @test length(model.constraints) == 22 || length(model.constraints) == 23
```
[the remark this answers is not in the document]
louis-gautier: Update: I tried different versions for this test but the number of constraints is never consistent. I will keep this test at its previous form.

louis-gautier (approved):
I think we are now ready to merge with master.

src/CP/valueselection/learning/utils.jl
```diff
@@ -145,17 +172,46 @@ end
...
+Used to work in `supervisedLeanedHeuristic.jl` to retrieve an action (which is the index in `state.possibleValuesIdx`) from a value, 
+and in `variableoutputcpnn.jl`. 
+
+"""
+function from_id_to_order(state::Union{DefaultTrajectoryState, BatchedDefaultTrajectoryState, HeterogeneousTrajectoryState, BatchedHeterogeneousTrajectoryState}, value_id::Int64; which_list="possibleValuesIdx", idx_in_batch=1)
+    @assert !isnothing(state.possibleValuesIdx)
+    if which_list == "possibleValuesIdx"
+        return findfirst(x->x == value_id, state.possibleValuesIdx)
+    elseif which_list == "allValuesIdx"
+        if typeof(state) == HeterogeneousTrajectoryState
```
3rdCore, line 200:
```suggestion
        if isa(state,HeterogeneousTrajectoryState)
```

src/RL/representation/default/defaultstaterepresentation.jl
```diff
@@ -112,7 +111,11 @@ Update the StateRepesentation according to its Type and Featurization.
 """
 function update_representation!(sr::DefaultStateRepresentation, model::CPModel, x::AbstractIntVar)
     update_features!(sr, model)
+    ncon = sr.cplayergraph.numberOfConstraints
+    nvar = sr.cplayergraph.numberOfVariables
+    sr.possibleValuesIdx = map(v -> indexFromCpVertex(sr.cplayergraph, ValueVertex(v)), collect(x.domain))
```
3rdCore, line 116: Did we specifically test this change ?

src/RL/representation/default/cp_layer/accessors.jl
```diff
@@ -137,14 +137,37 @@ Base.zero(::Type{CPLayerGraph}) = CPLayerGraph()
...
+                contovar[i, neighbor - ncon] = 1
+            end
+        elseif isa(node, ValueVertex)
+            neighbors = LightGraphs.outneighbors(g, i)
+            for neighbor in neighbors
+                valtovar[i - ncon - nvar, neighbor - ncon] = 1
+            end
+        end
+    end
+    return contovar, valtovar
```
3rdCore, line 172: Did we tested this ?

src/RL/representation/default/defaultstaterepresentation.jl
gostreap: I don't think we tested it. Maybe we should create an issue for each behavior we want to test, without postponing the merge?

src/RL/representation/default/cp_layer/accessors.jl
louis-gautier: Yes, we have unit tests for this in test/heterogeneousstaterepresentation.jl.

PR #233, opened 2022-05-24
Batched VariableOutputCPNN

ziadelassal:
This small PR enables VariableOutputCPNN to support batch processing.

### DIFF
--- a/src/RL/nn_structures/variableoutputcpnn.jl
+++ b/src/RL/nn_structures/variableoutputcpnn.jl
@@ -19,27 +19,130 @@ Flux.@functor VariableOutputCPNN
 
 wears_mask(s::VariableOutputCPNN) = false
 
-function (nn::VariableOutputCPNN)(state::GraphTrajectoryState)
+function (nn::VariableOutputCPNN)(state::BatchedDefaultTrajectoryState)
+    
+    variableIdx = state.variableIdx #Vector of size B
+    #println("variableIdx: ", variableIdx)
+    batchSize = length(variableIdx)
 
-    variableIdx = state.variableIdx
-    possibleValuesIdx = state.possibleValuesIdx
+    #println("batchSize: ", batchSize)
+    #println("allValuesIdx: ", state.allValuesIdx)
+    #display(state.allValuesIdx)
+    
+    possibleValuesIdx = [deepcopy(indexes) for indexes in state.possibleValuesIdx]
+    
+    @assert batchSize == length(possibleValuesIdx)
+    #println("possibleValuesIdx: ", possibleValuesIdx) 
+    #display(possibleValuesIdx)
+    actionSpaceSizes = [length(possibleValuesIdxPerVar) for possibleValuesIdxPerVar in possibleValuesIdx]
+    #println("actionSpaceSizes: ", actionSpaceSizes)
+    maxActionSpaceSize = Base.maximum(actionSpaceSizes) #the largest action space size in the batch
+    #println("maxActionSpaceSize: ", maxActionSpaceSize)
 
     # chain working on the graph(s)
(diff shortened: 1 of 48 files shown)


src/RL/nn_structures/variableoutputcpnn.jl
```diff
@@ -17,61 +17,42 @@ end
 # Enable the `|> gpu` syntax from Flux
 Flux.@functor VariableOutputCPNN
 
-wears_mask(s::VariableOutputCPNN) = false
```
3rdCore, line 20: Let's discuss this, I know that this condition is necessary to be able to handle action space of fixed size which is required for batch-learning. Otherwise, `variable_outputCPNN` was originally designed to address problems with a potentially large total_action space (union of all action space), so that unlike FFCPNN, the pipeline only computes the useful approximate Q-value for a fixed variable.

For the moment we don't solve problems that have this specificity,  so we can decide to keep that as it is, but that 'd make `VariableOutputCPNN` really close from `FFCPNN` (except the cross weight value/variable), That'd be nice to analyse the impact of these weights in the performance. 

what do you think fo this @ziadelassal ?

src/RL/nn_structures/variableoutputcpnn.jl
```diff
@@ -82,49 +63,37 @@ function (nn::VariableOutputCPNN)(state::BatchedDefaultTrajectoryState)
...
-            #println("singleFinalInput: ", size(singleFinalInput))
             finalInput = isempty(finalInput) ? [singleFinalInput] : append!(finalInput, [singleFinalInput])
         end
     end
-    #finalInput: vector of matrices of size F'xA (total size BxF'xA)
     Zygote.ignore() do
         f, a = size(finalInput[1])
-        finalInput = reshape(collect(Iterators.flatten(finalInput)), (f, maxActionSpaceSize, batchSize)) #!!TO TEST #convert vector of matrices into a 3-dimensional matrix
-        #println("finalInput: ", size(finalInput)) 
+        finalInput = reshape(collect(Iterators.flatten(finalInput)), (f, maxActionSpaceSize, batchSize)) #!!TO TEST #convert vector of matrices into a 3-dimensional matrix
```
3rdCore, line 87: > #!!TO TEST #convert vector of matrices into a 3-dimensional matrix 

Is it up-to-date ?

test/RL/nn_structures/heterogeneousvariableoutputcpnn.jl
```diff
@@ -0,0 +1,96 @@
...
+        
+        var = [1,2] #out of bound variable
+        val = [[1,2,3], [3]]
+        hts = SeaPearl.BatchedHeterogeneousTrajectoryState(fg, var, val)
+        bhts = hts |> cpu  |> cpu 
+        display(bhts.fg.valnf)
+        println(hts.possibleValuesIdx)
+        output = modelNN(bhts)
+        display(output)
+        @test size(output) == (size(valtovar,1), size(valtovar, 3))
```
3rdCore, line 82: This is linked to what is being discussed above. 

the approximate Q-values are computed for all the values, including the ones that don't belong to the variable domain. This was incompatible with _batch_learning_, which I should have mentioned before, I apologize.

3rdCore:
small comment : I suggest you to merge this branch to `heterogeneous` not to `master`. 

Otherwise, looks good for me after discussing the _wear_mask_ attribute.

src/RL/nn_structures/variableoutputcpnn.jl
ziadelassal: Also, we're working on a new CPNN that has cross weight value/variable, we could analyze the impact of these weights then

ziadelassal:
The branch has been merged on heterogeneous. Thus, I'm closing this PR.

PR #237, opened 2022-06-09
Correction of inconsistencies on GeneralReward

louis-gautier:
This small PR corrects two major problems on `GeneralReward`:

- Proper management of the first action which is taken: as the first pruning depends only on the instance itself, we should initialize the denominator of the variable part of the reward after it is taken (we were previously doing it during the construction of the reward).
- Proper management of the last action which is taken: reward from the `DecisionPhase` was previously overwritten by the reward given during the `EndingPhase`.


src/CP/valueselection/learning/rewards/generalreward.jl
```diff
@@ -69,9 +70,9 @@ function set_reward!(::Type{EndingPhase}, lh::LearnedHeuristic{SR, GeneralReward
 }
     # The rewards given in the EndingPhase ensure that any found feasible solution always gets a higher reward than any infeasible solution
     if symbol == :FoundSolution
-        lh.reward.value = 0
+        lh.reward.value += 0
     else
-        lh.reward.value = -lh.reward.gamma
+        lh.reward.value += -lh.reward.gamma
```
3rdCore, line 75: A few months ago, we did the same mistake. Good job finding this error, does this fix is related to the unexpected reward value we had for `latin` ?

louis-gautier: Both of the fixes in this PR help to fix the reward values in latin. I just have to look at the testsets that it broke and we will be able to merge it into `master`.

3rdCore (approved):

PR #239, opened 2022-06-15
New explorers and tripartite graph plot utils

louis-gautier:
This PR introduces a new explorer (softmax with temperature decay and UCB explorer) to test new exploring strategies for our RL agents.
It also adds a util which allows to plot tripartite graphs after they are initialized.


louis-gautier (commented):
I think that we can merge this branch into master now.

gostreap (approved):

PR #242, opened 2022-06-16
Specific state representations for graph coloring and MIS

louis-gautier:
This PR introduces some specific state representations for graph coloring and Maximum Independent Set to compare the performance they yield with those given by our tripartite graph state representation.


src/RL/representation/MIS/MISstaterepresentation.jl
```diff
@@ -0,0 +1,101 @@
+struct MISFeaturization <: AbstractFeaturization end
+
+"""
+    MISStateRepresentation{F, TS}
+
+This is a standard MIS representation using the graph of the problem with nodes having their assigned value as label.
+"""
+mutable struct MISStateRepresentation{F, TS} <: FeaturizedStateRepresentation{F, TS}
+    graph::LightGraphs.Graph
```
gostreap, line 9: It should have been great to use `Graphs.Graph` instead of `LightGraphs.Graph` since `LightGraphs` is deprecated and SeaPearl should transition towards `Graphs` in the long run.

I think that all the functions of `LightGraphs` used in this file are implemented in `Graphs` and that the work consists simply to replace `LightGraphs` with `Graphs`.

src/RL/representation/graphcoloring/graphcoloringstaterepresentation.jl
```diff
@@ -0,0 +1,105 @@
...
+Create nodeFeatures for every node of the graph (current color of the node or -1 if the color has not been determined yet)
+"""
+function featurize(sr::FeaturizedStateRepresentation{GraphColoringFeaturization, TS}) where TS
+    n = LightGraphs.nv(sr.graph)
+    nodeFeatures = zeros(Float32, 2, n)
+    for i in 1:n-1
+        nodeFeatures[1,i] = -1
+        nodeFeatures[2,i] = 0
+    end
+    nodeFeatures[1,n] = 0
```
gostreap, line 85: Why 0 instead of -1 here?

src/RL/representation/graphcoloring/graphcoloringstaterepresentation.jl
```diff
@@ -0,0 +1,105 @@
...
+    nodeFeatures[2,n] = 1
```
gostreap, line 86: Maybe put a small comment about the meaning of the second feature.

louis-gautier: Right, I will do it in an incoming commit.

src/RL/representation/graphcoloring/graphcoloringstaterepresentation.jl
louis-gautier: It was to indicate the agent that the objective had a different nature than regular nodes but it might be redundant with the other feature: I will remove it in an incoming commit.

PR #267, opened 2022-10-26
Update Project.toml to trigger a change of version

jardinetsouffleton:
Second attempt to use julia's registrator bot....


jardinetsouffleton:
@JuliaRegistrator register()

jardinetsouffleton:
Finally understood how to create new package version..... and it was not that

PR #322, opened 2023-01-30
Activity Based Value Selection Heuristic

3rdCore:
I implemented the ABS heuristic based on this paper https://arxiv.org/abs/1105.6314.

This PR is ready for review.

ToDos : 
- change alpha as an hyperparameters of the CPmodel (currently hardcoded in fixpoint.jl)


src/CP/core/model.jl
```diff
@@ -44,7 +44,7 @@ mutable struct Statistics
     objectiveUpPruning                      ::Union{Nothing, Float32}
     lastVar                                 ::Union{Nothing, AbstractIntVar} #last var on which we branched
     lastVal                                 ::Union{Nothing, Int} #last val on which we branched
-    searchTreeSize                          ::Union{Nothing, Int} #size of the search tree for impact based search
+    searchTreeSize                          ::Union{Nothing, StateObject{Int}} #size of the search tree for impact based search
```
3rdCore, line 47: `searchTreeSize` is now a `StateObject{Int}`. This makes the variable dependent of the model.trailer that ensure that the value is updated once the search backtrack in the searchTree.

3rdCore: Therefore, `searchTreeSize` has now to be updated using the function `setValue!`

PR #426, opened 2023-07-13
Rename abstract type pool to Pool

### LINKED ISSUE #425: pool is a type and a variable
`pool` is an [abstract type and a variable](https://github.com/corail-research/SeaPearl.jl/blob/master/src/RL/utils/geometricflux/heterogeneousgraphconv.jl). This leads to some bugs with later Julia versions.
```julia
function HeterogeneousGraphConv(ch::Pair{Int,Int}, original_dimensions::Array{Int}, σ=Flux.leakyrelu; init=Flux.glorot_uniform, bias::Bool=true, T::DataType=Float32, pool::pool=meanPooling())
```
I have fixed this by renaming the abstract type to `Pool`, as camel case is common for type names in Julia.

CasBex:
Closes #425

### DIFF
--- a/src/RL/utils/geometricflux/geometricflux.jl
+++ b/src/RL/utils/geometricflux/geometricflux.jl
@@ -1,11 +1,11 @@
 using NNlib: batched_mul
 
 
-abstract type pool end
+abstract type Pool end
 
-struct maxPooling <: pool end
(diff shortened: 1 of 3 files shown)


jardinetsouffleton:
Thank you very much @CasBex !!

